PR #22, opened 2022-04-22
Snapshot streaming for tessellation.

kriscieslak:


.scalafix.conf
```diff
@@ -0,0 +1,10 @@
+rules = [
+    OrganizeImports
+]
+
+OrganizeImports {
+  groupedImports                    = Merge
+  coalesceToWildcardImportThreshold = 3
+  groups                            = ["re:javax?\\.", "cats.", "scala.", "org.tessellation.", "*"]
+  removeUnused                      = false
```
marcinwadon, line 9: I'd set that to `true`

.scalafmt.conf
```diff
@@ -1,18 +1,16 @@
-version=2.0.0-RC5
+version="3.5.1"
+
+runner.dialect = scala3
 
 style = default
 project.git = true
 
 maxColumn = 120
-newlines.alwaysBeforeTopLevelStatements = true
```
marcinwadon, line 7: what about those 3 removed rules?

build.sbt
```diff
@@ -1,61 +1,92 @@
...
 
-version := "1.1.6"
-scalaVersion := "2.12.10"
-organization := "org.constellation"
+lazy val root = (project in file("."))
+  .settings(
+    name := "cl-snapshot-streaming",
+    testFrameworks += new TestFramework("weaver.framework.CatsEffect"),
+    libraryDependencies ++= catsDependencies ++ catsEffectDependencies ++ catsEffectTestDependencies ++ circeDependencies ++
+      configDependency ++ fs2Dependencies ++ http4sDependencies ++ loggingDepencencies ++ weaverDependency,
```
marcinwadon, line 13: I'd recommend the same pattern as we have in `tessellation` but that's up to you

src/main/resources/application.conf
```diff
@@ -1,31 +1,10 @@
...
-
-  mode {
-    getGenesis = false
-    getSnapshots = true
-  }
-  interval {
-    startingHeight = 2
-    endingHeight = 100000
+  last-ordinal-path = "/tmp/lastOrdinal.txt"
+  httpclient {
```
marcinwadon, line 3: what about `httpClient` or `http-client`?

src/main/scala/org/constellation/snapshotstreaming/App.scala
```diff
@@ -1,328 +1,66 @@
...
-  }
-}
+      nodeClient = NodeClient.make(client)
+      lastOrdinalService = LastOrdinalService.make(configuration)
+      snapshotService = SnapshotService.make(nodeClient, lastOrdinalService, configuration)
+
+      res <- snapshotService.getSnapshots(None).evalTap(s => Logger[F].info(s.toString))
+      _ <- saveLastOrdinalToFile(res.ordinal.value)
+    yield res
+    end for
```
marcinwadon, line 57: With scala3 and such for syntax, I'd suggest writing it like that, but it's up to you:
```scala
def mkClient =  Stream.resource(
         EmberClientBuilder
           .default[F]
           .withTimeout(configuration.httpClientTimeout)
           .withIdleTimeInPool(configuration.httpClientIdleTime)
           .build

val lastOrdinalService = LastOrdinalService.make(configuration)
          
mkClient.flatMap { client =>
  val nodeClient = NodeClient.make(client)
  val snapshotService = SnapshotService.make(nodeClient, lastOrdinalService, configuration)
  
  snapshotService.getSnapshots(None)
}
.flatTap { s => Logger[F].info(s.toString) }
.flatTap { saveLastOrdinalToFile(_.ordinal.value) )
```

src/main/scala/org/constellation/snapshotstreaming/Configuration.scala
```diff
@@ -1,66 +1,20 @@
...
-
-  val elasticsearchBalancesIndex: String =
-    elasticsearch.getString("indexes.balances")
-
-  val maxParallelRequests: Int = elasticsearch.getInt("maxParallelRequests")
-  val maxWaitQueueLimit: Int = elasticsearch.getInt("maxWaitQueueLimit")
-
-  val bucketRegion: String =
-    bucket.getString("region")
+  private val httpclient = config.getConfig("snapshot-streaming.httpclient")
```
marcinwadon, line 10: aa

src/main/scala/org/constellation/snapshotstreaming/HelloWorld.scala
```diff
@@ -0,0 +1,12 @@
+package org.constellation.snapshotstreaming
+
+import cats.effect.IO
+
+object HelloWorld:
```
marcinwadon, line 5: I think it should be removed, shouldn't it?

src/main/scala/org/constellation/snapshotstreaming/Main.scala
```diff
@@ -0,0 +1,10 @@
+package org.constellation.snapshotstreaming
+
+import cats.effect.{IO, IOApp}
+
+object Main extends IOApp.Simple:
```
marcinwadon, line 5: what about this main? :D

src/main/scala/org/constellation/snapshotstreaming/NodeClient.scala
```diff
@@ -0,0 +1,37 @@
...
+  def getSnapshot(ordinal: SnapshotOrdinal): F[Snapshot]
+end NodeClient
+
+object NodeClient:
+
+  def make[F[_]: Concurrent: KryoSerializer](client: Client[F]) = new NodeClient[F]:
+
+    def getLatestOrdinal =
+      import org.http4s.circe.CirceEntityCodec.circeEntityDecoder
+      val req = Request[F](Method.GET, uri"http://127.0.0.1:9010/global-snapshot/latest/ordinal")
```
marcinwadon, line 27: DRY (please extract the api url)

src/main/scala/org/constellation/snapshotstreaming/SnapshotService.scala
```diff
@@ -0,0 +1,55 @@
...
+import org.tessellation.dag.snapshot.GlobalSnapshot
+import org.tessellation.security.signature.Signed
+
+import fs2.io.file.{Files, Path}
+import fs2.{Stream, text}
+import org.typelevel.log4cats.Logger
+import org.typelevel.log4cats.slf4j.Slf4jLogger
+
+trait SnapshotService[F[_]]:
+  def getSnapshots(startingOrdinal: Option[Long]): Stream[F, Snapshot]
```
marcinwadon, line 18: shouldn't it be `Option[SnapshotOrdinal]`?

src/test/scala/org/constellation/snapshotstreaming/SnapshotServiceSuite.scala
```diff
@@ -0,0 +1,107 @@
...
+        case Some(savedOrdinal) => Stream.emit(savedOrdinal)
+      }
+
+  def mkNodeClient(ref: Ref[IO, Long]) = new NodeClient[IO]:
+    def getLatestOrdinal: IO[SnapshotOrdinal] = ref.get.map(SnapshotOrdinal(_))
+    def getSnapshot(ordinal: SnapshotOrdinal): IO[Snapshot] = IO.pure(Snapshot(ordinal))
+
+  def mkSnapshotService(lastOrdinalService: LastOrdinalService[IO], nodeClient: NodeClient[IO]) =
+    SnapshotService.make[IO](nodeClient, lastOrdinalService, configuration)
+  end mkSnapshotService
```
marcinwadon, line 34: please be consistent. either use `end` or no

marcinwadon (changes requested):

.scalafix.conf
kriscieslak: Unfortunately here is message when run OrganizeImports:
`ScalafixRunException: "OrganizeImports.removeUnused" is not supported on Scala 3 as the compiler is not providing enough information. Run the rule with "OrganizeImports.removeUnused" set to false to organize imports while keeping potentially unused imports.`

build.sbt
kpudlik: +1 for pattern from `tessellation`

project/Dependencies.scala
```diff
@@ -0,0 +1,89 @@
+// scalafmt: { rewrite.scala3.convertToNewSyntax = false, rewrite.scala3.removeOptionalBraces = no }
```
marcinwadon, line 1: as above

marcinwadon (changes requested):

project/Dependencies.scala
```diff
@@ -0,0 +1,89 @@
...
+    val organizeImports = "0.6.0"
+    val refined = "0.9.28"
+    val tessellation = "0.6.0"
+    val weaver = "0.7.11"
+  }
+
+  object Libraries {
+    def cats(artifact: String): ModuleID = "org.typelevel" %% s"cats-$artifact" % V.cats
+
+    def catsEffectDefn(artifact: String): ModuleID =
```
marcinwadon, line 28: why not just `catsEffect` like every other lib name?

src/main/scala/org/constellation/snapshotstreaming/ProcessedSnapshotsService.scala
```diff
@@ -0,0 +1,53 @@
+package org.constellation.snapshotstreaming
+
+import java.nio.file.NoSuchFileException
+
+import cats.effect.kernel.Async
```
marcinwadon, line 5: please import `Async` from `import cats.effect.Async`, not from `kernel`

src/main/scala/org/constellation/snapshotstreaming/opensearch/schema/AddressBalance.scala
```diff
@@ -0,0 +1,27 @@
...
+object AddressBalance {
+
+  implicit val dateEncoder: Encoder[Date] =
+    Encoder.encodeString.contramap(date => date.toInstant.toString)
+
+  implicit val snapshotEncoder: Encoder[AddressBalance] = deriveEncoder
+
+  implicit class AddressBalanceOps(balance: AddressBalance) {
+    def docId = s"${balance.address}${balance.snapshotOrdinal}"
+  }
```
marcinwadon, line 25: you can simplify this as a function in the case class:
```
case class AddressBalance(
   address: String,
   balance: Long,
   snapshotHash: String,
   snapshotOrdinal: Long,
   timestamp: Date
 ) {
  def docId = s"${balance.address}${balance.snapshotOrdinal}"
 }
```
you don't need an implicit class for this

kpudlik (changes requested):

kpudlik (approved):

build.sbt
```diff
@@ -1,61 +1,66 @@
...
-  "com.github.scopt" %% "scopt" % versions.scopt,
-  ("org.constellation" %% "cl-schema" % versions.constellation).from(
-    s"https://github.com/Constellation-Labs/constellation/releases/download/v${versions.constellation}/cl-schema.jar"
-  ),
-  "com.sksamuel.elastic4s" %% "elastic4s-circe" % "6.7.8",
-  "com.amazonaws" % "aws-java-sdk-s3" % "1.11.863"
-)
+lazy val versions = new {
+  val weaver = "0.7.11"
+}
```
marcinwadon, line 62: Why `weaver` is treated differently that all the other libraries? It should be defined via `Libraries.weaver` as well

marcinwadon: Especially that this part (version) is duplicated

src/main/resources/application.conf
```diff
@@ -1,30 +1,23 @@
...
+  httpClient {
+    timeout = "120s"
+    idleTimeInPool = "30s"
   }
-  bucket {
-    urls = ["a","b","c"]
-    region = "us-west-1"
-    skipHeightOnFailure = false
+  node {
+    urls = ["http://13.52.246.74:9000/global-snapshot/"]
```
marcinwadon, line 8: I think it shouldn't be hardcoded here in example (see previous implementation)

src/main/scala/org/constellation/snapshotstreaming/App.scala
```diff
@@ -1,328 +1,45 @@
...
-  ): Stream[F, S3GenesisDeserializedResult] = {
-    s3DAOs match {
-      case List(bucket) =>
-        bucket.getGenesis().handleErrorWith { e =>
-          Stream.eval(
-            LiftIO[F].liftIO(
-              logger.error(e)(s"[S3-Genesis ->] ERROR. While getting Genesis from bucket=${bucket.getBucketName}.")
-            )
-          ) >> Stream.raiseError(e)
+    Slf4jLogger.create[IO].flatMap { implicit logger =>
```
marcinwadon, line 19: I left a comment here previously but looks like it has been ignored. Slf4jLogger creation part is duplicated (see: def mkLogger)

src/main/scala/org/constellation/snapshotstreaming/node/NodeDownload.scala
```diff
@@ -0,0 +1,56 @@
...
+      _ <- Stream.sleep(config.nodeIntervalInSeconds.seconds)
+      ordinal <- Stream
+        .eval(nodeClient.getLatestOrdinal)
+        .handleErrorWith(ex =>
+          Stream.eval(
+            logger.error(ex)(s"Couldn't download latest ordinal error from ${nodeClient.uri}.")
+          ) >> Stream.empty
+        )
+        .evalTap(ordinal => logger.debug(s"Latest ordinal from ${nodeClient.uri} is $ordinal."))
+    } yield ordinal).repeat
```
marcinwadon, line 36: I think that slow requests may slow down the whole stream. Instead of `Stream.sleep` I'd consider `awakeEvery` and `metered` eventually 🤔

src/main/scala/org/constellation/snapshotstreaming/node/NodeDownload.scala
```diff
@@ -0,0 +1,56 @@
...
+        .evalTap(s => logger.info(s"Snapshot ${s.value.ordinal.value} downloaded from ${nodeClient.uri}."))
+        .handleErrorWith(ex =>
+          Stream.eval(logger.error(ex)(s"Couldn't download snapshot $ordinal from ${nodeClient.uri}.")) >> Stream
+            .raiseError(ex)
+        )
+        .attempt
+        .map[Either[SnapshotOrdinal, Signed[GlobalSnapshot]]] {
+          case Left(_)  => Left(ordinal)
+          case Right(s) => Right(s)
+        }
```
marcinwadon, line 52: `.map(_.leftMap(_ => ordinal))`

src/main/scala/org/constellation/snapshotstreaming/node/SnapshotOrdinal.scala
```diff
@@ -0,0 +1,21 @@
...
+
+object SnapshotOrdinal {
+
+  implicit class LongToSnapshot(snapshotsOrdinalValue: Long) {
+    @inline def toSnapshotOrdinal: SnapshotOrdinal = SnapshotOrdinal(snapshotsOrdinalValue)
+  }
+
+  implicit class SnapshotOrdinalOps(snapshotOrdinal: SnapshotOrdinal) {
+    @inline def increment: SnapshotOrdinal = (snapshotOrdinal.value + 1).toSnapshotOrdinal
+  }
```
marcinwadon, line 16: as commented previously, you can write this directly in a case class

src/main/scala/org/constellation/snapshotstreaming/opensearch/ElasticSearchDAO.scala
```diff
@@ -0,0 +1,56 @@
...
+  def make[F[_]: Async](openSearchUrl: Uri) =
+    new ElasticSearchDAO[F] {
+
+      private val logger = Slf4jLogger.getLoggerFromClass[F](ElasticSearchDAO.getClass)
+
+      def getEsClient(): Resource[F, ElasticClient] = {
+        val client = ElasticClient(
+          JavaClient(ElasticProperties(s"${openSearchUrl.renderString}"))
+        )
+        Resource.fromAutoCloseable(logger.info("Initiating es client.") >> Async[F].delay(client))
```
marcinwadon, line 32: elasticsearch or opensearch?

PR #24, opened 2022-05-31
Remove not needed code. Change implementation of getting snapshot ordinal.

kriscieslak:


src/main/scala/org/constellation/snapshotstreaming/App.scala
```diff
@@ -12,19 +11,17 @@ import org.http4s.ember.client.EmberClientBuilder
 import org.typelevel.log4cats.slf4j.Slf4jLogger
 
 object App extends IOApp {
-  def mkLogger = Slf4jLogger.create[IO]
+  private val logger = Slf4jLogger.getLogger[IO]
```
marcinwadon, line 14: the previous implementation with `Slf4jLogger.create[IO]` was also correct, but then `mkLogger` was unused (could be just removed). anyway, both ways are correct imo

src/main/scala/org/constellation/snapshotstreaming/node/NodeDownload.scala
```diff
@@ -23,8 +24,8 @@ object NodeDownload {
 
     private val logger = Slf4jLogger.getLoggerFromClass[F](NodeDownload.getClass)
 
-    override def downloadLatestOrdinal(): Stream[F, SnapshotOrdinal] = (for {
-      _ <- Stream.sleep(config.nodeIntervalInSeconds.seconds)
+    override def downloadLatestOrdinal(): Stream[F, SnapshotOrdinal] = for {
+      _ <- Stream.awakeEvery(config.nodeIntervalInSeconds.seconds)
```
marcinwadon, line 28: have you tested that? how it works without `metered`?

kriscieslak: I have written unit test, though haven't run it yet. I will check it.

marcinwadon: my original comment related to that part was about long execution and I cannot see such a unit test 🤔 so I think it's untested part still

kriscieslak: I have added two more test cases in unit test. 
First, where action after request have long execution time, but no longer than interval. In this case request for next ordinal happens after defined interval - execution time of action has no impact on planned request.
Second, where action after request have long execution time and is longer than interval. In this case request for next ordinal happens after action finishes. 
In previous implementation next request had to wait for action to finish plus interval time.
Is it desired behaviour?
Do you think further tests should be done?

kriscieslak: I have tested it with local cluster with 3 nodes. It works. 
If execution time is long, but in the meantime only one snapshot is created, it checks for latest snapshot every given interval. When execution time is long and in the meantime more than one snapshot is created than in next iteration all created snapshot ordinals will be processed and after then request for latest ordinal will be sent.

kriscieslak: I have tested with dampen flag set to false (`Stream.awakeEvery(config.nodeIntervalInSeconds.seconds, dampen = false)`) and in such behaviour you mentioned can be observed. Otherwise requests for next ordinal are not queued.

marcinwadon: So it means that this is a wrong implementation, as both ways requests should not be queued 🤔 If the execution time will be longer, it can lead to memory overflow

kriscieslak: The `dampen` flag is set to `true` in implementation so we shouldn't bother what happens when it is set to `false`. I did just to check how it works. Unless I miss sth I can't see possiblity to have memory overflow in this implementation.

marcinwadon: I don't talk about the test but about the code implementation 🤔

PR #26, opened 2022-06-10
Add scripts for restarting service and cleaning indices.

kriscieslak:
…ndices.


terraform/templates/restart
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+sudo systemctl stop snapshot-streaming
+
```
marcinwadon, line 4: I'd execute clean_indices here between stop and start

terraform/templates/clean_indices.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
marcinwadon, line 1: why `start` and `restart` are without `.sh` ext and `clean_indices` is with it?

marcinwadon (changes requested):

.gitignore
```diff
@@ -10,3 +10,4 @@ logs/
 *.tfvars
 *.tfstate
 .terraform.lock.hcl
+*.jar
```
marcinwadon, line 13: Why?

terraform/ec2.tf
```diff
@@ -80,9 +81,23 @@ resource "aws_instance" "snapshot-streaming" {
...
+
+  provisioner "file" {
+    source = "templates/restart"
+    destination = "/home/ec2-user/snapshot-streaming/restart"
+  }
+
   provisioner "remote-exec" {
     inline = [
-      "sudo chmod 774 /home/ec2-user/snapshot-streaming/start",
+      "sudo chmod 774 /home/ec2-user/snapshot-streaming/start.sh",
```
marcinwadon, line 98: Why?

.gitignore
kriscieslak: Removed.

terraform/ec2.tf
marcinwadon: There were already two files, so you could add the third one in the same pattern

kriscieslak: Which pattern do you think is better? Should I change it back?

marcinwadon: Adding the new file I'd keep the existing pattern (and we use it across all projects and nodes)

terraform/templates/restart
```diff
@@ -0,0 +1,14 @@
...
+cat <<EOF > ./nextOrdinal.json
+{
+  "startingOrdinal": 0,
+  "gaps": []
+}
+EOF
+
+./clean_indices.sh
```
marcinwadon, line 12: It won't work now

marcinwadon (changes requested):

terraform/templates/snapshot-streaming.service
```diff
@@ -4,7 +4,7 @@ Description=Constellation Snapshot-Streaming
 [Service]
 User=ec2-user
 WorkingDirectory=/home/ec2-user/snapshot-streaming
-ExecStart=/home/ec2-user/snapshot-streaming/start
+ExecStart=/home/ec2-user/snapshot-streaming/start.sh
```
tbekas, line 7: In `terraform/ec2.tf` in line 98 the name of the script is provided without the `.sh` extension.
```
sudo chmod 774 /home/ec2-user/snapshot-streaming/start
```

tbekas (approved):
LGTM

marcinwadon (approved):

PR #38, opened 2022-10-19
Check if snapshots chain is preserved before uploading

TheMMaciek:


src/main/scala/org/constellation/snapshotstreaming/App.scala
```diff
@@ -1,39 +1,30 @@
...
-          .map(_ => ExitCode.Success)
-          .handleErrorWith(e => logger.warn(e)(e.getMessage).map(_ => ExitCode.Error))
+    Random.scalaUtilRandom[IO].asResource.use { implicit random =>
+      KryoSerializer.forAsync[IO](dag.dagSharedKryoRegistrar ++ shared.sharedKryoRegistrar).use { implicit ks =>
+        SecurityProvider.forAsync[IO].use { implicit sp =>
+          SnapshotProcessor.make[IO](configuration).use { snapshotProcessor =>
+            snapshotProcessor.runtime.compile.drain
+              .flatTap(_ => logger.debug("Done!"))
+              .map(_ => ExitCode.Success)
+              .handleErrorWith(e => logger.error(e)(e.getMessage).map(_ => ExitCode.Error))
```
marcinwadon, line 25:
```
.map(_ => A)
```
can be:
```
.as(A)
```

src/main/scala/org/constellation/snapshotstreaming/Configuration.scala
```diff
@@ -14,13 +22,24 @@ class Configuration {
...
-
-  val httpClientTimeout = Duration(httpClient.getString("timeout"))
-  val httpClientIdleTime = Duration(httpClient.getString("idleTimeInPool"))
-
-  val nodeIntervalInSeconds: Int = node.getInt("retryIntervalInSeconds")
-  val nodeUrls: List[Uri] = node.getStringList("urls").asScala.toList.map(Uri.unsafeFromString)
+  val lastSnapshotPath: Path = Path(config.getString("snapshotStreaming.lastSnapshotPath"))
+  val l0Peers: NonEmptyMap[PeerId, L0Peer] = NonEmptyMap.fromMapUnsafe(
+    SortedMap.from(
+      node.getStringList("l0Peers").asScala.toList.map(decode[L0Peer](_).toOption).map(p => p.get.id -> p.get)
```
marcinwadon, line 28: `p.get` may fail

src/main/scala/org/constellation/snapshotstreaming/FileBasedLastGlobalSnapshotStorage.scala
```diff
@@ -0,0 +1,82 @@
+package org.constellation.snapshotstreaming
+
+import java.nio.file.NoSuchFileException
+import cats.effect.Async
+import cats.effect.std.Semaphore
+import cats.syntax.applicativeError._
+import cats.syntax.flatMap._
+import cats.syntax.functor._
```
marcinwadon, line 8: do we have a scalafmt/fix here, like in tessellation?

src/main/scala/org/constellation/snapshotstreaming/FileBasedLastGlobalSnapshotStorage.scala
```diff
@@ -0,0 +1,82 @@
...
+      implicit val codec: Codec[Hashed[GlobalSnapshot]] = deriveCodec[Hashed[GlobalSnapshot]]
+
+      private val logger = Slf4jLogger.getLogger[F]
+
+      def set(snapshot: Hashed[GlobalSnapshot]): F[Unit] =
+        semaphore.permit.use { _ =>
+          get.flatMap {
+            case Some(last) if isNextSnapshot(last, snapshot) =>
+              Stream
+                .emit(snapshot.asJson.spaces2)
```
marcinwadon, line 41: why we need `spaces2` if we're not going to read this file manually?

src/main/scala/org/constellation/snapshotstreaming/SnapshotProcessor.scala
```diff
@@ -0,0 +1,141 @@
...
+
+    private def process(snapshot: Hashed[GlobalSnapshot]): F[Unit] =
+      lastGlobalSnapshotStorage.get.flatMap {
+        case Some(last) if Validator.isNextSnapshot(last, snapshot) =>
+          s3DAO.uploadSnapshot(snapshot) >>
+            prepareAndExecuteBulkUpdate(snapshot) >>
+            lastGlobalSnapshotStorage.set(snapshot)
+
+        case Some(last) =>
+          logger.warn(s"Pulled snapshot doesn't form a correct chain, ignoring! Last: ${GlobalSnapshotReference
```
marcinwadon, line 87: will it try fetching from another node?

marcinwadon (changes requested):

src/main/scala/org/constellation/snapshotstreaming/FileBasedLastGlobalSnapshotStorage.scala
TheMMaciek: We may look into it in some cases and it doesn't break anything.

src/main/scala/org/constellation/snapshotstreaming/FileBasedLastGlobalSnapshotStorage.scala
TheMMaciek: We have something but it seems not completely configured. I wanted to create a ticket to configure it as a separate task and not do it here to not mix issues in one PR.

src/main/scala/org/constellation/snapshotstreaming/Configuration.scala
TheMMaciek: And I want it to fail if it's not decoded well. But I will move the `get` to right after the `decode`. Anyway we could improve the whole configuration reading and starting part, but that would require changing the library I guess and I didn't want to mix it in into this PR. I can create a low prio ticket for that.

src/main/scala/org/constellation/snapshotstreaming/SnapshotProcessor.scala
TheMMaciek: Yes it works similarly to how it works in dag L1. The pulling will execute again in couple of seconds (configured with `pullInterval`) and it will try to fetch the next snapshots after the last accepted one. Wrong snapshots are not accepted, so the retry will happen from a randomly picked node.

src/main/scala/org/constellation/snapshotstreaming/SnapshotProcessor.scala
```diff
@@ -0,0 +1,141 @@
...
+              .fromHashedGlobalSnapshot(last)} pulled: ${GlobalSnapshotReference.fromHashedGlobalSnapshot(snapshot)}")
+
+        case None if configuration.initialSnapshot.exists(initial => initial.hash === snapshot.hash && initial.ordinal === snapshot.ordinal) =>
+          s3DAO.uploadSnapshot(snapshot) >>
```
TheMMaciek, line 91: Should we handle the situation when one of these fail and just retry again on another pull? The question is how will s3 or opensearch behave when the file already exists or partial update to indexes has been already made? @tbekas @kpudlik @kriscieslak @marcinwadon

src/main/scala/org/constellation/snapshotstreaming/FileBasedLastGlobalSnapshotStorage.scala
marcinwadon: we can use `jq` then

TheMMaciek: We can. I can address that if you want. Personally I don't see that as a problem or being worse/better.

src/main/scala/org/constellation/snapshotstreaming/SnapshotProcessor.scala
TheMMaciek: Going to handle the failure and retry again.

marcinwadon: One of these you mean s3 succeed but opensearch not? imo we should retry. we should definitely not process next ordinals

kriscieslak (approved):
LGTM

src/main/scala/org/constellation/snapshotstreaming/SnapshotProcessor.scala
TheMMaciek: Yeah I meant if one of these fails, should we pull the same snapshot again and try to upload to s3/opensearch that might have the snapshot uploaded or partially uploaded already. Anyway I tested it locally with dockerized opensearch and localstack (mimicks aws) and resending doesn't throw any errors so we can retry. One thing that could happen with retry like this is that if at this particular ordinal the cluster forked and created two snapshots we may on the subsequent tries download any of these snapshots and hence upload parts/everything of both to s3/opensearch. Though this scenario seems to be of little probability. Before handling the error here the whole snapshot-streaming app would stop.

marcinwadon (approved):

PR #39, opened 2022-10-31
Load whole snapshot file before deserializing

TheMMaciek:


src/main/scala/org/constellation/snapshotstreaming/FileBasedLastGlobalSnapshotStorage.scala
```diff
@@ -64,11 +66,12 @@ object FileBasedLastGlobalSnapshotStorage {
...
           .through(text.utf8.decode)
-          .map(decode[Hashed[GlobalSnapshot]])
-          .map(_.toOption)
           .compile
-          .last
-          .map(_.flatten)
+          .toList
+          .map(_.mkString)
+          .map(decode[Hashed[GlobalSnapshot]])
+          .flatMap(_.liftTo[F])
```
marcinwadon, line 73: isn't `.flatMap(_.liftTo[F])` equal to `.map(a => a)` which is just an identity?

marcinwadon (approved):

PR #42, opened 2023-02-14
Skaffold pipeline and dependency version updates

tbekas:


.github/workflows/release.yml
```diff
@@ -9,31 +9,51 @@ name: Create Release
...
         with:
-          java-version: 1.8
-      - name: Assemble assets
-        run: |
-          sbt "++ 2.12.10" assembly
+          distribution: temurin
+          java-version: 11
+          cache: sbt
+
+      - name: Check sbt version
```
tbekas, line 34: This got committed by accident. It will be removed with the code-review fixes.

.github/workflows/pull_request.yml
```diff
@@ -5,16 +5,40 @@ name: Pull request tests.
...
-        uses: actions/setup-java@v1
+      - name: Set up Java
+        uses: actions/setup-java@v3
         with:
-          java-version: 1.8
+          distribution: temurin
+          java-version: 11
+          cache: sbt
+
+      - name: Check sbt version
```
tbekas, line 30: I committed this step by accident. I will remove it along with code-review fixes.

.github/workflows/release.yml
tbekas: Fixed

.github/workflows/pull_request.yml
tbekas: Fixed

.github/workflows/pull_request.yml
```diff
@@ -5,16 +5,35 @@ name: Pull request tests.
...
+        uses: actions/checkout@v3
+        with:
+          path: snapshot-streaming
+
+      - name: Checkout tessellation
+        uses: actions/checkout@v3
+        with:
+          path: tessellation
+          repository: Constellation-Labs/tessellation
+          ref: v1.9.1
```
TheMMaciek, line 20: I think we need to scrape it from `project/Dependencies.scala` file, right?

.github/workflows/pull_request.yml
```diff
@@ -5,16 +5,35 @@ name: Pull request tests.
...
-        uses: actions/setup-java@v1
+      - name: Set up Java
+        uses: actions/setup-java@v3
         with:
-          java-version: 1.8
+          distribution: temurin
+          java-version: 11
+          cache: sbt
+
+      - name: Install tessellation
```
TheMMaciek, line 30: Maybe `Build tessellation packages`?

.github/workflows/release.yml
```diff
@@ -9,31 +9,46 @@ name: Create Release
...
+
+      - name: Set up Java
+        uses: actions/setup-java@v3
         with:
-          java-version: 1.8
-      - name: Assemble assets
-        run: |
-          sbt "++ 2.12.10" assembly
+          distribution: temurin
+          java-version: 11
```
TheMMaciek, line 31: Do we want to stick with java 11 rather than 8?

README.md
```diff
@@ -1 +1,33 @@
-# cl-snapshot-streaming
\ No newline at end of file
+# Snapshot Streaming
+
+## Building
+
+In the root directory of [tessellation](https://github.com/cngo-github/tessellation) run
+
+```
+git checkout v1.9.1
```
TheMMaciek, line 8: As well here we should checkout and build the version from `project/Dependencies.scala`

.github/workflows/pull_request.yml
tbekas: Well, I can change this and I thought about doing it right away, but I decided against it. My reasoning is that this is a _temporary_ solution and I think we shouldn't have any sophisticated temporary solutions. We should have permanent solutions - and that is publishing packages to GitHub packages repository.

build.sbt
```diff
@@ -1,13 +1,20 @@
...
-ThisBuild / version := "2.0.0"
+ThisBuild / scalaVersion := "2.13.10"
 ThisBuild / scalafixDependencies += Libraries.organizeImports
-ThisBuild / assemblyMergeStrategy in assembly := {
-  case PathList("META-INF", xs @ _*)            => MergeStrategy.discard
-  case PathList("org", "joda", "time", xs @ _*) => MergeStrategy.last
-  case x                                        => MergeStrategy.first
+ThisBuild / evictionErrorLevel := Level.Warn
+
+val ghTokenSource = TokenSource.GitConfig("github.token") || TokenSource.Environment("GITHUB_TOKEN")
```
TheMMaciek, line 8: Maybe just assign straight to `githubTokenSource` without this intermediary val? It's not reused anywhere if I checked correctly.

.github/workflows/release.yml
tbekas: Well, I don't think we should stick with any Java version. I think we should just advance it to the newer LTS versions that are compatible with our software whenever we find it suitable (jdk 1.8 was released almost a decade ago). I know that testnet and mainnet instances of snapshot-streaming are currently running jdk 1.8, but I think we can update them to java 11 as well.

build.sbt
tbekas: Makes sense.

README.md
tbekas: As above, I think we should avoid sophisticated temporary solutions.

TheMMaciek (changes requested):
Looks good overall. One thing I think may needs improving is scraping the version of tessellation.

TheMMaciek (approved):
We are not changing the scraping logic for the tessellation so I thinks it looks ok.

PR #43, opened 2023-02-16
Updated release flow

tbekas:


doc/release-flow.md
```diff
@@ -0,0 +1,6 @@
+# Release flow
+
+1. Checkout the `develop` branch and pull the latest changes
```
kpudlik, line 3: Are we going to stick with simplified (`develop` only) flow? I mean in other repos we have `main` + `release/vA.B.C` flow

tbekas: This is my proposition. I think that `snapshot-streaming` doesn't have much of an active development currently, so in my opinion there's no need to have two branches `main` and `develop`. Let's simplify our work whenever we can.

kpudlik (approved):

PR #45, opened 2023-03-16
Use incremental snapshots

kriscieslak:


src/main/resources/application.conf
```diff
@@ -1,5 +1,6 @@
 snapshotStreaming {
   lastSnapshotPath = "lastSnapshot.json"
+  lastIncrementalSnapshotPath = "lastIncrementalSnapshot.json"
   httpClient {
```
TheMMaciek, line 9: I guess we also need config for collateral here.

src/main/scala/org/constellation/snapshotstreaming/FileBasedLastGlobalSnapshotStorage.scala
```diff
@@ -1,84 +1,94 @@
...
-      def setInitial(snapshot: Hashed[GlobalSnapshot]): F[Unit] =
-        semaphore.permit.use { _ =>
-          Stream
-            .emit(snapshot.asJson.spaces2)
-            .through(text.utf8.encode)
-            .through(Files[F].writeAll(path, Flags(Flag.Write, Flag.CreateNew)))
-            .compile.drain
-        }
+  def make[F[_]: Async](path: Path) = new FileBasedLastGlobalSnapshotStorage[F] {
+    implicit val codec: Codec[Hashed[GlobalSnapshot]] = deriveCodec[Hashed[GlobalSnapshot]]
```
TheMMaciek, line 75: I guess if you are not saving or reading the `Hashed[GlobalSnapshot]` you also don't need a codec for it.

src/main/scala/org/constellation/snapshotstreaming/GlobalSnapshotReader.scala
```diff
@@ -0,0 +1,64 @@
...
+object GlobalSnapshotReader {
+
+  def make[F[_]: Async](
+    globalSnapshotClient: L0GlobalSnapshotClient[F],
+    l0ClusterStorage: L0ClusterStorage[F],
+    snapshotStorage: FileBasedLastGlobalSnapshotStorage[F]
+  ): GlobalSnapshotReader[F] = new GlobalSnapshotReader[F] {
+    private val logger = Slf4jLogger.getLogger[F]
+
+    implicit val codec: Codec[Hashed[GlobalSnapshot]] = deriveCodec[Hashed[GlobalSnapshot]]
```
TheMMaciek: We can extract it if it's reused in multiple places in snapshot-streaming to not redefine it in separate classes. The thing about no codec for `Hashed` in tessellation is that it's not supposed to be sent over network, it should be for internal usage only. No supplied codec kind of helps for it to be that way.

src/main/scala/org/constellation/snapshotstreaming/GlobalSnapshotReader.scala
```diff
@@ -0,0 +1,64 @@
...
+
+      def getSnapshot = getIncremental
+        .map(_.asRight[Signed[GlobalSnapshot]])
+        .orElse(getFull.map(_.asLeft[Signed[IncrementalGlobalSnapshot]]))
+
+      getSnapshot.getOrRaise(new Throwable("Cannot find neither GlobalSnapshot nor IncrementalGlobalSnapshot."))
+
+    }
+
+    private def pullGlobalSnapshot(hash: Hash): F[Option[Signed[IncrementalGlobalSnapshot]]] =
```
TheMMaciek: Maybe shorter `pullIncremental`

src/main/scala/org/constellation/snapshotstreaming/SnapshotProcessor.scala
```diff
@@ -84,25 +127,26 @@ object SnapshotProcessor {
...
+        case Some(last) if Validator.isNextSnapshot(last, snapshot.signed.value) =>
           s3DAO.uploadSnapshot(snapshot) >>
-            prepareAndExecuteBulkUpdate(snapshot) >>
-            lastGlobalSnapshotStorage.set(snapshot)
+            prepareAndExecuteBulkUpdate(snapshot, snapshotInfo) >>
+            lastIncrementalGlobalSnapshotStorage.set(snapshot, snapshotInfo)
 
         case Some(last) =>
           logger.warn(
-            s"Pulled snapshot doesn't form a correct chain, ignoring! Last: ${getSnapshotReference(last)} pulled: ${getSnapshotReference(snapshot)}"
```
TheMMaciek, line 96: Why not import in a shortened format and use it that way?

src/main/scala/org/constellation/snapshotstreaming/SnapshotProcessor.scala
```diff
@@ -115,46 +159,94 @@ object SnapshotProcessor {
...
         .awakeEvery(configuration.pullInterval)
         .evalMap { _ =>
-          lastGlobalSnapshotStorage.get.flatMap { last =>
+          lastIncrementalGlobalSnapshotStorage.getCombined.flatMap { last =>
             (last, configuration.initialSnapshot) match {
-              case (Some(_), _) =>
+              case (Some((lastSnapshot, lastState)), _) =>
                 l0Service.pullGlobalSnapshots
+                  .map(
+                    _.leftMap(_ => new Throwable(s"!"))
```
TheMMaciek: Could be more descriptive 😄

src/main/scala/org/constellation/snapshotstreaming/FileBasedLastFullGlobalSnapshotStorage.scala
```diff
@@ -0,0 +1,56 @@
...
+trait FileBasedLastFullGlobalSnapshotStorage[F[_]] {
+  def get: F[Option[Signed[GlobalSnapshot]]]
+}
+
+object FileBasedLastFullGlobalSnapshotStorage {
+
+  def make[F[_]: Async](path: Path) = new FileBasedLastFullGlobalSnapshotStorage[F] {
+    implicit val codec: Codec[Hashed[GlobalSnapshot]] = deriveCodec[Hashed[GlobalSnapshot]]
+
+    def set(snapshot: Hashed[GlobalSnapshot]): F[Unit] =
```
TheMMaciek, line 31: Not used anywhere, possibly that's why it's not a part of the interface as well.

TheMMaciek: This should be a part of the interface. It still works and can be called right now because `make` method has a complex derived type of:
```
FileBasedLastFullGlobalSnapshotStorage[IO] with Object {
  val codec : Codec[Hashed[GlobalSnapshot]]
  def set(snapshot: Hashed[GlobalSnapshot]): IO[Unit]
}
```
I would also recommend to declare the return type for `make`.

src/main/scala/org/constellation/snapshotstreaming/FileBasedLastIncrementalGlobalSnapshotStorage.scala
```diff
@@ -0,0 +1,121 @@
...
+      implicit val codec: Codec[Hashed[GlobalIncrementalSnapshot]] = deriveCodec[Hashed[GlobalIncrementalSnapshot]]
+      implicit val snapshotWithInfoCodec: Codec[SnapshotWithState] = deriveCodec[SnapshotWithState]
+
+      def set(snapshot: Hashed[GlobalIncrementalSnapshot], state: GlobalSnapshotInfo): F[Unit] =
+        semaphore.permit.use { _ =>
+          validateStateProof(snapshot, state) >>
+            get.flatMap {
+              case Some(last) if isNextSnapshot(last, snapshot.signed.value) =>
+                Stream
+                  .emit(SnapshotWithState(snapshot, state).asJson.spaces2)
```
TheMMaciek, line 53: I used `spaces2` when I first implemented the file based storage, to have the file readable without any additional tools, but we could also write it down with `noSpaces` and just pipe through `jq` if readable output is needed. Up to you.

src/main/scala/org/constellation/snapshotstreaming/FileBasedLastIncrementalGlobalSnapshotStorage.scala
```diff
@@ -0,0 +1,121 @@
...
+              .through(Files[F].writeAll(path, Flags(Flag.Write, Flag.CreateNew)))
+              .compile
+              .drain
+        }
+
+      private def validateStateProof(snapshot: Hashed[GlobalIncrementalSnapshot], state: GlobalSnapshotInfo) =
+        state
+          .stateProof[F]
+          .map(_ === snapshot.stateProof)
+          .flatMap((new Throwable("State proof doesn't match!").raiseError[F, Unit].unlessA))
```
TheMMaciek: I understand that this is a critical error, right? A snapshot with invalid state proof should never make it here unless there would be an implementation error? I wonder about whether we should throw or not.

src/main/scala/org/constellation/snapshotstreaming/SnapshotProcessor.scala
```diff
@@ -49,33 +52,54 @@ object SnapshotProcessor {
...
+    l0Service: GlobalL0Service[F],
     opensearchDAO: OpensearchDAO[F],
     s3DAO: S3DAO[F],
-    updateRequestBuilder: UpdateRequestBuilder[F]
+    updateRequestBuilder: UpdateRequestBuilder[F],
+    tessellationServices: TessellationServices[F],
+    lastFullGlobalSnapshotStorage: FileBasedLastFullGlobalSnapshotStorage[F]
   ): SnapshotProcessor[F] = new SnapshotProcessor[F] {
-
+    FileBasedLastFullGlobalSnapshotStorage
```
TheMMaciek: Some leftover I guess.

TheMMaciek (approved):

PR #49, opened 2023-03-29
Update tessellation v2.0.0 alpha.2

kriscieslak:


.github/workflows/release.yml
```diff
@@ -21,7 +21,7 @@ jobs:
         with:
           path: tessellation
           repository: Constellation-Labs/tessellation
-          ref: v2.0.0-alpha.1
+          ref: v2.0.0-alpha.2
```
TheMMaciek, line 24: You need to change it in dependencies as well.

TheMMaciek (approved):

PR #51, opened 2023-06-13
Add currency snapshots processing

TheMMaciek:


.github/workflows/pull_request.yml
```diff
@@ -17,7 +17,7 @@ jobs:
         with:
           path: tessellation
           repository: Constellation-Labs/tessellation
-          ref: v2.0.0-alpha.6
+          ref: process-currency-snapshots-returns-all-processed-currency-snapshots
```
TheMMaciek: To be removed for the final non-draft version. We need to release the tessellation first and then adjust the dependency here and in Dependencies.scala

kpudlik (dismissed):

marcinwadon (dismissed):

.github/workflows/pull_request.yml
TheMMaciek: Adjusted

kpudlik (dismissed):

kpudlik (approved):

PR #59, opened 2024-03-13
Update tessellation to 2.4.0-SNAPSHOT

marcinwadon:


.envrc
```diff
@@ -0,0 +1 @@
+use flake
```
TheMMaciek, line 1: Should that be a part of the PR?

src/main/resources/application.conf
```diff
@@ -1,6 +1,7 @@
 snapshotStreaming {
   lastSnapshotPath = "lastSnapshot.json"
   lastIncrementalSnapshotPath = "lastIncrementalSnapshot.json"
+  environment = Testnet
```
TheMMaciek, line 4: that's the default config?

.envrc
marcinwadon: yep :+1:

src/main/scala/org/constellation/snapshotstreaming/FileBasedLastIncrementalGlobalSnapshotStorage.scala
```diff
@@ -29,15 +29,17 @@ import io.circe.syntax._
...
+  def make[F[_]: Async: Files: KryoSerializer: Hasher](
+    path: Path,
+    hashSelect: HashSelect
   ): F[LastSnapshotStorage[F, GlobalIncrementalSnapshot, GlobalSnapshotInfo]] =
     Semaphore[F](1)
-      .map(make(path, _))
+      .map(make(path, _, hashSelect))
 
-  private def make[F[_]: Async: Files: KryoSerializer](
+  private def make[F[_]: Async: Files: KryoSerializer: Hasher](
```
TheMMaciek, line 39: Is `KryoSerializer` still needed here given you added Hasher?

src/main/resources/application.conf
marcinwadon: that's the example config, as far as I can see:
```
  node {
    l0Peers = ["""{"id": "<id>", "ip": "<ip>", "port": "<port>"}"""]
    pullInterval = "5s"
    pullLimit = 4
    # terminalSnapshotOrdinal = 30
  }
  opensearch {
    host = "http://sample"
    port = "80"
```

src/main/scala/org/constellation/snapshotstreaming/FileBasedLastIncrementalGlobalSnapshotStorage.scala
marcinwadon: No idea. Probably I'd have to change the build parameters as you did in tessellation, to see that. Right?

TheMMaciek: Or just try removing it and see if it compiles.

TheMMaciek:
LGTM

TheMMaciek (approved):

src/main/scala/org/constellation/snapshotstreaming/FileBasedLastIncrementalGlobalSnapshotStorage.scala
marcinwadon: Ok, I'll try that in a separate PR

PR #61, opened 2024-05-13
Add FeeTransaction handling for balances

kpudlik:


project/Dependencies.scala
```diff
@@ -18,7 +18,7 @@ object Dependencies {
     val logstash = "7.2"
     val organizeImports = "0.6.0"
     val refined = "0.10.1"
-    val tessellation = "2.5.0"
+    val tessellation = "2.7.0-SNAPSHOT"
```
kpudlik: We need to set it to the next release after `2.6.2`

sdavidp (approved):

src/test/scala/org/constellation/snapshotstreaming/CurrencySnapshotMapperSuite.scala
```diff
@@ -0,0 +1,274 @@
...
+        feeTransactions = feeTransactions
+      )
+
+      result = CurrencyIncrementalSnapshotMapper.make().snapshotReferredBalancesInfo(snapshot, totalInfo)
+      expectedBalances = createBalances(address1, address2, address3, address4, address5, address7)
+    } yield expect.same(result, expectedBalances)
+  }
+
+  test(
+    "leave balances for addresses from transactions and fee transactions, but not set zero for destinations not in balances "
```
TheMMaciek, line 160: How is it that two tests with the same name pass? 🤔

kpudlik: Because we have some common test cases for both Global and Currency mappers

PR #62, opened 2024-05-28
Add fee, stakingAddress, ownerAddress and sizeInKB to CurrencySnapshot

kpudlik:


src/main/scala/org/constellation/snapshotstreaming/TessellationServices.scala
```diff
@@ -3,46 +3,67 @@ package org.constellation.snapshotstreaming
...
+        CurrencySnapshotCreator.make[F](currencySnapshotAcceptanceManager, None, configuration.snapshotSize, currencyEventsCutter)
+      }
+      currencySnapshotValidator = {
+        implicit val hs = alwaysCurrentHasherSelector
+        CurrencySnapshotValidator.make[F](currencySnapshotCreator, SignedValidator.make[F], None, None)
+      }
+      currencySnapshotContextFns = {
+        implicit val hs = alwaysCurrentHasherSelector
+        CurrencySnapshotContextFunctions.make(currencySnapshotValidator)
+      }
```
sdavidp: Couldn't we make `alwaysCurrentHasherSelector` implicit? Then we don't have to create the implicit 3 times in blocks

sdavidp: Or, if that's going to conflict with `hashSelector` later, use one block to create all three variables?

sdavidp: I just tried the above and maybe doesn't read as well. WDYT?

src/main/scala/org/constellation/snapshotstreaming/opensearch/mapper/GlobalSnapshotMapper.scala
```diff
@@ -1,39 +1,84 @@
...
+        val transactions = snapshot.blocks.flatMap(_.block.transactions.toSortedSet)
+        val source = transactions.map(_.source)
+        val destination = transactions.map(_.destination)
+        SnapshotReferredAddresses(source, destination)
+      }
+
+      def mapSnapshot(snapshot: Hashed[GlobalIncrementalSnapshot], timestamp: Date, hasher: Hasher[F]): F[Snapshot] =
+        for {
+          blocksHashes <- snapshot.blocks.unsorted.map(_.block).map(hashBlock(_, hasher)).toList.sequence
+        } yield Snapshot(
```
sdavidp:
```suggestion
        snapshot.blocks.unsorted.map(_.block).map(hashBlock(_, hasher))
          .toList
          .sequence
          .map(blockHashes => Snapshot(

sdavidp (dismissed):

src/main/scala/org/constellation/snapshotstreaming/TessellationServices.scala
kpudlik: We only care about creating `currencySnapshotContextFns` so I am fine with wrapping in a single block :+1:

kpudlik: Oh, actually we can even get rid of implicit current hasher selector for currency completely because currency uses Json only so there is no need to select hasher based on ordinal.

src/main/scala/org/constellation/snapshotstreaming/opensearch/mapper/GlobalSnapshotMapper.scala
kpudlik: fixed

marcinwadon (dismissed):

marcinwadon (approved):

PR #70, opened 2024-12-03
feat: parallelizing sending bulk updates to OpenSearch on same snapshot and updating tessellation version

marcus-girolneto:
### Changes
+ To address recent issues with snapshot streaming when sending data to OpenSearch, the bulk update process has been optimized by introducing parallelization.
+ Now, updates for each snapshot are sent in parallel while ensuring that only one snapshot is processed at a time.
+ This optimization helped improve the performance on 2024-12-02, helping the system catch up with the network.

### Tests

Screenshot 2024-12-03 at 10 38 35

### NOTE
**I've generated a JAR with these changes from version 4.3.3 and deployed it to the SnapshotStreaming of Mainnet, @AlexBrandes authorized this since we needed to do something to catch up with the network, so we need to generate an official version and maybe re-deploy**

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -18,7 +18,7 @@ object Dependencies {
     val logstash = "7.2"
     val organizeImports = "0.6.0"
     val refined = "0.10.1"
-    val tessellation = "2.7.1"
+    val tessellation = "2.12.0"
     val weaver = "0.8.1"
   }
 
--- a/src/main/scala/org/constellation/snapshotstreaming/SnapshotProcessor.scala
+++ b/src/main/scala/org/constellation/snapshotstreaming/SnapshotProcessor.scala
@@ -1,12 +1,13 @@
 package org.constellation.snapshotstreaming
 
 import java.util.Date
-import cats.Applicative
+import cats.{Applicative, Parallel}
 import cats.data.NonEmptyList
 import cats.data.NonEmptyMap
 import cats.data.Validated
 import cats.effect._
 import cats.effect.std.Random
+import cats.effect.syntax.all._
 import cats.syntax.all._
 import org.tessellation.currency.schema.currency.CurrencyIncrementalSnapshot
 import org.tessellation.currency.schema.currency.CurrencySnapshot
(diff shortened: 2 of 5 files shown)


src/main/scala/org/constellation/snapshotstreaming/SnapshotProcessor.scala
```diff
@@ -119,7 +119,18 @@ object SnapshotProcessor {
...
+            for {
+              start <- Clock[F].realTime
+              _ <- logger.info(s"Starting to send bulk updates to Opensearch")
+              result <- requests.parTraverse { br =>
+                opensearchDAO.sendToOpensearch(bulk(br))
+              }
+              end <- Clock[F].realTime
+              elapsedTime = end.toMillis - start.toMillis
+              _ <- logger.info(s"Bulk update operation took $elapsedTime ms")
+            } yield result
```
scasplte2: I believe cats provides a method for timing that does this for you

``` scala
logger.info(s"Starting to send bulk updates to Opensearch") >>
requests.parTraverse { br =>
  opensearchDAO.sendToOpensearch(bulk(br))
}.timed.flatTap { case (elapsedTime, _) =>
  logger.info(s"Bulk update operation took ${elapsedTime.toMillis} ms")
}
```

project/Dependencies.scala
```diff
@@ -76,7 +76,7 @@ object Dependencies {
 
     val organizeImports = "com.github.liancheng" %% "organize-imports" % V.organizeImports
 
-    val tessellationSdk = "org.constellation" %% "tessellation-sdk" % V.tessellation
+    val tessellationNodeShared = "org.constellation" %% "tessellation-node-shared" % V.tessellation
```
scasplte2: Why the switch? I thought Marcin was urging us to move to only using `sdk` and this change doesn't use any additional tooling from Tessellation.

src/main/scala/org/constellation/snapshotstreaming/SnapshotProcessor.scala
marcus-girolneto: 👍🏼

project/Dependencies.scala
marcus-girolneto: 👍🏼

scasplte2 (approved):
LGTM

PR #71, opened 2024-12-17
Add postgres dev

gclaramunt:


sql/snapshot.sql
```diff
@@ -0,0 +1,100 @@
...
+-- DROP TABLE global_snapshot;
+
+CREATE TABLE global_snapshot (
+	ordinal int8 NOT NULL,
+	hash text NOT NULL,
+	height int8 NOT NULL,
+	subheight int8 NOT NULL,
+	last_snapshot_hash text NULL,
+	created_at timestamp NULL,
+	updated_at timestamp NULL,
```
marcinwadon: why we need `updated_at` field? afaik all data here are immutable and once stored, never updated 🤔

sql/snapshot.sql
```diff
@@ -0,0 +1,100 @@
...
+CREATE TABLE dag_transaction (
+	hash text NOT NULL,
+	global_ordinal int8 NULL,
+	block text NULL,
+	"source" text NOT NULL,
+	destination text NOT NULL,
+	amount int8 NOT NULL,
+	fee int8 NOT NULL,
+	block_hash text NOT NULL,
+	transaction_original_id int4 NULL,
```
marcinwadon: what is this field? 🤔

sql/snapshot.sql
```diff
@@ -0,0 +1,100 @@
+-- public.global_snapshot definition
+
+-- Drop table
+
+-- DROP TABLE global_snapshot;
+
+CREATE TABLE global_snapshot (
```
AlexBrandes: Notes: 
- `hash` can be VARCHAR
- `subheight` can be int4
- `last_snapshot_hash` can be VARCHAR
- We need to include fields from the current snapshot schema here as nullable fields:
    - `epochProgress`
    - `metagraphSnapshotCount`
- We should have a unique index/constraint on `hash`
- `last_snapshot_hash` should FK to `hash` on this table
- `created_at` / `updated_at` - need DEFAULT NOW() to set initial value and then I think updated_at needs a trigger to update whenever it’s changed. We should consider timestampz instead of timestamp here but maybe it doesn’t matter.

sql/snapshot.sql
```diff
@@ -0,0 +1,100 @@
...
+CREATE INDEX global_snapshot_hash_idx ON public.global_snapshot USING btree (hash);
+
+
+-- public.dag_balance_change definition
+
+-- Drop table
+
+-- DROP TABLE dag_balance_change;
+
+CREATE TABLE dag_balance_change (
```
AlexBrandes: Notes: 
- `ordinal` - since we have global ordinals and metagraph/currency ordinals I would suggest making each explicit every time we use it. This would become `global_ordinal`.
- `address` can be VARCHAR
- This will be queried by `address` and then `ordinal` frequently so we probably need an index on `(address, ordinal)` w/address first. I think we can just make that the order of the PK and that would work.
- Do we want to have an `addresses` table that this FKs to? We’ll have several tables that reference `address`.
- Should we add `created_at` and `updated_at` to all tables? Or what’s the thought behind which tables get those columns and which don’t?

sql/snapshot.sql
```diff
@@ -0,0 +1,100 @@
...
+);
+
+
+-- public.dag_transaction definition
+
+-- Drop table
+
+-- DROP TABLE dag_transaction;
+
+CREATE TABLE dag_transaction (
```
AlexBrandes: Notes: 
- `hash` can be VARCHAR
- `global_ordinal` - nice, let’s use this format throughout. Need an index on this column to fetch transactions per snapshot.
- `block` - is this supposed to be `block_hash`? Can be VARCHAR.
- `source` - this is a wallet address. Should be VARCHAR.
- `destination` - also a wallet address
- `block_hash` - what’s the difference between this and `block`?
- `transaction_original_id` - does this FK to a separate table? I’m not really sure why we return this from the API but we’ll need to continue returning it for backwards compatibility. We could store it as JSON here as it’s never queried separately from this record and it’s a nested record so it would mean multiple tables to store it separately.
- `created_at` / `updated_at` - same as above

sql/snapshot.sql
```diff
@@ -0,0 +1,100 @@
...
+);
+
+
+-- public.metagraph_snapshot definition
+
+-- Drop table
+
+-- DROP TABLE metagraph_snapshot;
+
+CREATE TABLE metagraph_snapshot (
```
AlexBrandes: Notes: 
- `metagraph_id` should be VARCHAR
- `ordinal` - should we make this `metagraph_ordinal` ? Or just `ordinal` for metagraph snapshots and `global_ordinal` for global snapshots? We should be consistent on all tables.
- `global_ordinal` - I think this should be NOT NULL. Is there a case this could be NULL?
- `hash` - VARCHAR
- `subheight` - probably int4 since this doesn’t get big
- `last_snapshot_hash` VARCHAR
- `created_at` / `updated_at` - same as above
- global_snapshot_fk constraint references the metagraph ordinal but we need 2 FKs here anyways: global snapshot and metagraph snapshot.
- Need `epochProgress` here also and probably other CurrencySnapshot fields (`fee` definitely). Take a look at the output in an actual snapshot.

src/main/scala/org/constellation/snapshotstreaming/Configuration.scala
```diff
@@ -87,4 +87,17 @@ class Configuration(sharedConfigReader: SharedConfigReader) {
...
+  import pureconfig.ConfigSource
+  import pureconfig.generic.auto._
+
+  case class DbConfig(host: String, port: Int, user: String, password: Option[String], database: String, maxSessions: Int)
+
+  lazy val dbConfig: DbConfig = ConfigSource
+    .default
+    .at("snapshotStreaming.db")
+    .loadOrThrow[DbConfig]
+
```
marcinwadon: can we use the pattern that we already have configured in the project? it would be great not to make two different ways of reading config, so either use the previous one or refactor the previous one to use pureconfig instead (but as a separate PR) 🤔

sql/snapshot.sql
```diff
@@ -0,0 +1,100 @@
...
+);
+
+
+-- public.metagraph_transaction definition
+
+-- Drop table
+
+-- DROP TABLE metagraph_transaction;
+
+CREATE TABLE metagraph_transaction (
```
AlexBrandes: Notes: 
- `hash` VARCHAR
- `metagraph_id` VARCHAR
- `block` - what is this?
- `source` / `destination` - same as above about addresses
- `transaction_original_id` - same as above
- `created_at` / `updated_at` - same as above

build.sbt
```diff
@@ -64,6 +64,8 @@ lazy val core = (project in file("."))
       Libraries.fs2Core,
       Libraries.fs2Io,
       Libraries.guava,
+      Libraries.skunk,
```
marcinwadon, line 68: what about doobie? 🤔

AlexBrandes (changes requested):
**Overall Notes:** 
- Table names should be plural (global_snapshots, dag_balance_changes, etc.)
- Think about what indexes will be needed based on the existing endpoint definitions. 
- Tables that still need to be added:
- - metagraph balances
- - blocks
- - DAG rewards + metagraph rewards
- - FeeTransactions
- - Metagraphs

sql/snapshot.sql
AlexBrandes: That's not necessarily true. OpenSearch indexes are edited sometimes in unusual cases, for example. The only overhead for `updated_at` is the trigger when it's actually updated so it doesn't really impact performance significantly. It makes the database a lot more human readable in case of any issues though. 

I prefer to have them on all tables but it's up for debate if there's a good reason to remove those columns.

marcinwadon: > OpenSearch indexes are edited sometimes in unusual cases, for example.

@AlexBrandes could you elaborate on that? what is an example case?

sql/snapshot.sql
AlexBrandes: There are a few other missing columns here if you check against the existing BE API output: salt, parent, etc.

https://be-mainnet.constellationnetwork.io/transactions?limit=10

sql/snapshot.sql
AlexBrandes: You would know better than I would but haven't you guys had to edit or reindex in certain cases? There have been the sporadic issues with balances (I know some of those were issues between 2 indexes). There was also an older case I remember that had a broken chain that had to be fixed. 

I can think of other cases when we might update a row instead of deleting it and re-inserting also. For example, we might decide that we really need all historical state proofs from all snapshots to be returned through the API. Updating the existing rows would be a better strategy than deleting/re-inserting everything. 

I guess the important questions are:
- Will this field impact performance enough that we should avoid it? imo no
- Will this field be useful enough for working with the DB that it's worth including? imo yes

marcinwadon: AFAIK all cases were that the data were not indexed. Since we ingest blockchain data that are immutable, we should never update data indexed in the database. (For example, an exchange confirmed a transaction and then you update its balance or parent hash, or whatever). AFAIK block explorer, since it's centralized, works as a source of truth for the whole network. Cluster nodes fetch data from block explorer when performing a rollback. Imo those data should never be modified.

gclaramunt: yeah, I was wondering the same, it was in the initial draft in notion and seemed reasonable :P I'll take them out.

src/main/scala/org/constellation/snapshotstreaming/Configuration.scala
gclaramunt: This is a WIP and the main goal is to validate the approach and library. Once it reaches a mature state, I'll unify it with the rest of the project. I favor PureConfig because is less boilerplate.

build.sbt
gclaramunt: TL;DR: "use Doobie for general-purpose database interaction and multi-database support and use Skunk if you're committed to PostgreSQL and want the most type-safe, performant, and PostgreSQL-optimized solution."

I think is worth a shot for our use case. 
(anyway, since the syntax/api usage is pretty similar, switching later is not that big impact)

https://www.notion.so/SQL-libraries-160a94bd4b4f8011b72bfb43caa0fea8

src/main/scala/org/constellation/snapshotstreaming/Configuration.scala
marcinwadon: Ok. Me too, but EOD I'd opt for consistency so either this or that

gclaramunt:
regarding varchar vs text, I don't have a preference, the underlying datatype is the same. Since size is fixed, should we consider char(SIZE) ?

sql/snapshot.sql
AlexBrandes: So it would be a rare occurrence that a record would be updated - in general I agree. 

What's the argument against including the column though? Performance?

AlexBrandes:
> regarding varchar vs text, I don't have a preference, the underlying datatype is the same. Since size is fixed, should we consider char(SIZE) ?

Ah, I didn't know this. That's not the case in MySQL, for example. 

I think TEXT is fine then. I feel like we will run into edge cases we can't support if we force an exact length with char(SIZE).

sql/snapshot.sql
gclaramunt: Both arguments make sense. Although adding the `updated_at` column later is not free, if we're careful, we can minimize the impact, so I don't think we need to commit right now :) 
(since updates are rare, maybe we can consider a separate table also)

src/main/scala/org/constellation/snapshotstreaming/opensearch/UpdateRequestBuilder.scala
```diff
@@ -10,89 +10,117 @@ import com.sksamuel.elastic4s.requests.update.UpdateRequest
...
+          blocks,
+          transactions,
+          balances,
+          currSnapshot,
+          currIncrementalSnapshots,
+          currBlocks,
+          currTransactions,
+          currFeeTransactions,
+          currBalances
+        ).grouped(config.bulkSize).toList
```
marcinwadon: maybe you can put whole split snapshot into the `updateParallelRequests` function?

src/main/scala/org/constellation/snapshotstreaming/opensearch/UpdateRequestBuilder.scala
```diff
@@ -10,89 +10,117 @@ import com.sksamuel.elastic4s.requests.update.UpdateRequest
...
+        currencySnapshots,
+        maybePrevSnapshotInfo.map(_.lastCurrencySnapshots),
+        timestamp,
+        txHasher,
+        hasher
+      )
+      (currSnapshot, currIncrementalSnapshots, currBlocks, currTransactions, currFeeTransactions, currBalances) =
+        mappedCurrencyData
+    } yield {
+      val (snapshot, blocks, transactions, balances) = mappedGlobalData
```
marcinwadon: why currency data are in the for comp and global data are in yield? imo only the SplitSnapshot should be in the yield. or even, create a fn in the SplitSnapshot companion object that will receive mapped global data and mapped currency data, and produce the split snapshot. then, that whole for comprehension may look like:
```scala
(globalMapper.mapGlobalSnapshot(...), currencyMapper.mapCurrencySnapshots(...))
  .mapN(SplitSnapshot(_, _))
```

src/main/scala/org/constellation/snapshotstreaming/opensearch/UpdateRequestBuilder.scala
```diff
@@ -10,89 +10,117 @@ import com.sksamuel.elastic4s.requests.update.UpdateRequest
...
 import org.constellation.snapshotstreaming.Configuration
+import org.constellation.snapshotstreaming.opensearch.SplitSnapshot.splitAndHash
+import org.constellation.snapshotstreaming.schema.{AddressBalance, Block, CurrencyData, CurrencySnapshot, FeeTransaction, Snapshot, Transaction}
 
 case class UpdateRequests(
   sequentialRequests: Seq[Seq[UpdateRequest]],
   parallelRequests  : List[List[UpdateRequest]]
 )
 
+case class SplitSnapshot(snapshot: Snapshot, blocks: Seq[Block], txs: List[Transaction], balances: Seq[AddressBalance], cdSnapshots: Seq[CurrencyData[Snapshot]], cdCcySnapshots: Seq[CurrencyData[CurrencySnapshot]], cdBlocks: Seq[CurrencyData[Block]], cdTxs: Seq[CurrencyData[Transaction]], cdFeeTxs: Seq[CurrencyData[FeeTransaction]], cdBalances: Seq[CurrencyData[AddressBalance]])
```
marcinwadon: what is the use case for the `SplitSnapshot`? 🤔 I see you create the class and then desctructure it immediately after 🤔

sql/snapshot.sql
AlexBrandes: Still need to add `epochProgress` - also add `version` and `proofs` as a separate relation.

sql/snapshot.sql
```diff
@@ -9,9 +30,8 @@ CREATE TABLE block (
 	hash varchar NOT NULL,
 	height int8 NOT NULL,
 	parent jsonb NULL,
-	lock_timestamp timestamp NOT NULL,
-	snapshot_hash varchar NOT NULL,
-	snapshot_ordinal int8 NOT NULL,
+	created_at timestamp DEFAULT now() NOT NULL,
+	updated_at timestamp DEFAULT now() NOT NULL,
```
AlexBrandes, line 43: I'll just comment here but it applies to all the tables. You'll need an auto-update trigger to support `updated_at`.

sql/snapshot.sql
```diff
@@ -23,16 +43,38 @@ CREATE TABLE block (
...
 	height int8 NOT NULL,
-	subheight int8 NOT NULL,
+	subheight int4 NOT NULL,
 	last_snapshot_hash varchar NOT NULL,
-	created_at timestamp NOT NULL DEFAULT NOW(),
-	updated_at timestamp NOT NULL DEFAULT NOW(),
-	CONSTRAINT global_snapshot_pkey PRIMARY KEY (ordinal)
+	created_at timestamp DEFAULT now() NOT NULL,
+	updated_at timestamp DEFAULT now() NOT NULL,
+	CONSTRAINT global_snapshot_pk PRIMARY KEY (hash)
```
AlexBrandes: `ordinal` needs to be a unique constraint

sql/snapshot.sql
```diff
@@ -23,16 +43,38 @@ CREATE TABLE block (
...
+	hash varchar NOT NULL,
+	height int8 NOT NULL,
+	subheight int4 NOT NULL,
+	last_snapshot_hash varchar NULL,
+	fee int8 NOT NULL,
+	owner_address varchar NULL,
+	staking_address varchar NULL,
+	created_at timestamp DEFAULT now() NOT NULL,
+	updated_at timestamp DEFAULT now() NOT NULL,
+	CONSTRAINT metagraph_snapshot_pk PRIMARY KEY (metagraph_id, hash)
```
AlexBrandes: `(metagraph_id, ordinal)` should be a unique constraint here also.

sql/snapshot.sql
```diff
@@ -42,13 +84,13 @@ CREATE INDEX global_snapshot_hash_idx ON public.global_snapshot USING btree (has
 -- DROP TABLE dag_balance_change;
 
 CREATE TABLE dag_balance_change (
-    hash varchar NOT NULL,
-	ordinal int8 NOT NULL,
+	snapshot_hash varchar NOT NULL,
```
AlexBrandes, line 191: Even if `snapshot_hash` is the FK to snapshots here, we need `global_ordinal` because we need to be able to query this by address and ordinal like this

```
SELECT balance 
FROM dag_balance_changes 
WHERE global_ordinal < :snapshot_ordinal and address = :address 
LIMIT 1
```

sql/snapshot.sql
```diff
@@ -42,13 +84,13 @@ CREATE INDEX global_snapshot_hash_idx ON public.global_snapshot USING btree (has
...
 	address varchar NOT NULL,
 	balance int8 NULL,
-	ts timestamp NOT NULL,
-	CONSTRAINT dag_balance_change_pk PRIMARY KEY (hash, ordinal, address),
-	CONSTRAINT dag_balance_change_global_snapshot_fk FOREIGN KEY (ordinal) REFERENCES global_snapshot(ordinal)
+	created_at timestamp DEFAULT now() NOT NULL,
+	updated_at timestamp DEFAULT now() NOT NULL,
+	CONSTRAINT dag_balance_change_pk PRIMARY KEY (snapshot_hash, address),
```
AlexBrandes:
```suggestion
	CONSTRAINT dag_balance_change_pk PRIMARY KEY (address, global_ordinal),
```

sql/snapshot.sql
```diff
@@ -60,85 +102,83 @@ CREATE TABLE dag_balance_change (
 
 CREATE TABLE dag_block (
 	hash varchar NOT NULL,
-	global_snapshot_ordinal int8 NULL,
-	CONSTRAINT dag_block_pkey PRIMARY KEY (hash),
+	snapshot_hash varchar NOT NULL,
+	created_at timestamp DEFAULT now() NOT NULL,
+	updated_at timestamp DEFAULT now() NOT NULL,
+	CONSTRAINT dag_block_pk PRIMARY KEY (hash),
 	CONSTRAINT dag_block_block_fk FOREIGN KEY (hash) REFERENCES block(hash),
```
AlexBrandes: Is this an FK to the parent abstract_table? I think we can probably remove.

sql/snapshot.sql
```diff
@@ -60,85 +102,83 @@ CREATE TABLE dag_balance_change (
...
 
--- DROP TABLE dag_transaction;
+-- DROP TABLE dag_reward_transaction;
 
-CREATE TABLE dag_transaction (
-	hash varchar NOT NULL,
-	block varchar NULL,
-	source_addr varchar NOT NULL,
+CREATE TABLE dag_reward_transaction (
 	destination_addr varchar NOT NULL,
```
AlexBrandes, line 245: I would call this one `destination` to match the transaction schema

sql/snapshot.sql
```diff
@@ -60,85 +102,83 @@ CREATE TABLE dag_balance_change (
...
-	block_hash varchar NOT NULL,
-	original jsonb NULL,
-	created_at timestamp NOT NULL DEFAULT NOW(),
-	updated_at timestamp NOT NULL DEFAULT NOW(),
-	CONSTRAINT dag_transaction_pkey PRIMARY KEY (hash),
-	CONSTRAINT dag_transaction_dag_block_fk FOREIGN KEY (block) REFERENCES dag_block(hash)
+	global_snapshot_hash varchar NOT NULL,
+	created_at timestamp DEFAULT now() NOT NULL,
+	updated_at timestamp DEFAULT now() NOT NULL,
+	CONSTRAINT reward_transaction_global_snapshot_fk FOREIGN KEY (global_snapshot_hash) REFERENCES global_snapshot(hash)
```
AlexBrandes: This table has no PK.

sql/snapshot.sql
AlexBrandes: Still missing the columns in the above comment.

AlexBrandes (commented):
- I think we need an addresses table that all the different transaction tables FK to (source/destination).

sql/snapshot.sql
```diff
@@ -0,0 +1,251 @@
...
+
+-- DROP TABLE dag_transaction;
+
+CREATE TABLE dag_transaction (
+	hash varchar NOT NULL,
+	block_hash varchar NOT NULL,
+	created_at timestamp DEFAULT now() NOT NULL,
+	updated_at timestamp DEFAULT now() NOT NULL,
+	CONSTRAINT dag_transaction_pk PRIMARY KEY (hash),
+	CONSTRAINT dag_transaction_abstract_transaction_fk FOREIGN KEY (hash) REFERENCES abstract_transaction(hash),
```
AlexBrandes: Does it need an FK to the abstract table? I'm not familiar with how that works.

sql/snapshot.sql
```diff
@@ -0,0 +1,251 @@
...
+	hash varchar NOT NULL,
+	amount int8 NOT NULL,
+	source_addr varchar NOT NULL,
+	destination_addr varchar NOT NULL,
+	parent jsonb NOT NULL,
+	metagraph_snapshot_hash varchar NULL,
+	metagraph_id varchar NULL,
+	created_at timestamp DEFAULT now() NOT NULL,
+	updated_at timestamp DEFAULT now() NOT NULL,
+	CONSTRAINT fee_transaction_pkey PRIMARY KEY (hash),
```
AlexBrandes: I think this needs to be at least `(metagraph_id, hash)`. We should double check that the hash is unique to the metagraph also. It's generated from the data but I don't remember how it's made unique.

sql/snapshot.sql
```diff
@@ -0,0 +1,251 @@
...
+-- DROP TABLE fee_transaction;
+
+CREATE TABLE fee_transaction (
+	hash varchar NOT NULL,
+	amount int8 NOT NULL,
+	source_addr varchar NOT NULL,
+	destination_addr varchar NOT NULL,
+	parent jsonb NOT NULL,
+	metagraph_snapshot_hash varchar NULL,
+	metagraph_id varchar NULL,
```
AlexBrandes:
```suggestion
	metagraph_snapshot_hash varchar NOT NULL,
	metagraph_id varchar NOT NULL,
```

sql/snapshot.sql
```diff
@@ -0,0 +1,251 @@
...
+
+-- public.metagraph_balance_change definition
+
+-- Drop table
+
+-- DROP TABLE metagraph_balance_change;
+
+CREATE TABLE metagraph_balance_change (
+	metagraph_id varchar NOT NULL,
+	metagraph_hash varchar NOT NULL,
```
AlexBrandes: Same issue as the `dag_balance_change` table. We need ordinal here to be queried, unless the intent is to join to the metagraph_snapshots table. That seems like it might be too inefficient though to search against.

sql/snapshot.sql
```diff
@@ -0,0 +1,251 @@
...
+	address varchar NOT NULL,
+	balance int8 NULL,
+	created_at timestamp DEFAULT now() NOT NULL,
+	updated_at timestamp DEFAULT now() NOT NULL,
+	CONSTRAINT metagraph_balance_change_metagraph_snapshot_fk FOREIGN KEY (metagraph_id,metagraph_hash) REFERENCES metagraph_snapshot(metagraph_id,hash)
```
AlexBrandes: This table has no PK

sql/snapshot.sql
```diff
@@ -0,0 +1,251 @@
...
+);
+
+
+-- public.metagraph_block definition
+
+-- Drop table
+
+-- DROP TABLE metagraph_block;
+
+CREATE TABLE metagraph_block (
```
AlexBrandes: This table has no PK.

sql/snapshot.sql
```diff
@@ -0,0 +1,251 @@
...
+);
+
+
+-- public.metagraph_reward_transaction definition
+
+-- Drop table
+
+-- DROP TABLE metagraph_reward_transaction;
+
+CREATE TABLE metagraph_reward_transaction (
```
AlexBrandes: This table has no PK.

sql/snapshot.sql
```diff
@@ -0,0 +1,251 @@
...
+);
+
+
+-- public.metagraph_snapshots_to_global_snapshot definition
+
+-- Drop table
+
+-- DROP TABLE metagraph_snapshots_to_global_snapshot;
+
+CREATE TABLE metagraph_snapshots_to_global_snapshot (
```
AlexBrandes: I think this works - metagraph_id and metagraph_snapshot_hash have to be NOT NULL and the FK to metagraph_snapshot would need to be removed because these rows would be inserted before the snapshot records exist. This table does feel a little bit messy though if it doesn't have FK constraints. 

It might be easier to just to the `metagraph_snapshot_count` or a JSONB field with an array of metagraph snapshot records on the global snapshot. They would only be used while upload of the metagraph snapshots is pending.

sql/snapshot.sql
```diff
@@ -0,0 +1,251 @@
...
+	global_snapshot_hash varchar NOT NULL,
+	metagraph_snapshot_hash varchar NOT NULL,
+	metagraph_id varchar NULL,
+	created_at timestamp DEFAULT now() NOT NULL,
+	updated_at timestamp DEFAULT now() NOT NULL,
+	CONSTRAINT metagraph_snapshots_to_global_snapshot_global_snapshot_fk FOREIGN KEY (global_snapshot_hash) REFERENCES global_snapshot(hash),
+	CONSTRAINT metagraph_snapshots_to_global_snapshot_metagraph_snapshot_fk FOREIGN KEY (metagraph_id,metagraph_snapshot_hash) REFERENCES metagraph_snapshot(metagraph_id,hash)
```
AlexBrandes: This table has no PK.

AlexBrandes (commented):
Sorry - hit submit on the last review too quick. Here are a few more comments.

sql/snapshot.sql
```diff
@@ -0,0 +1,546 @@
+-- public.abstract_blocks definition
+
+-- Drop table
+
+-- DROP TABLE abstract_blocks;
+
+CREATE TABLE abstract_blocks (
+	hash varchar NOT NULL,
```
kpudlik, line 10: We can use `varchar(64)` for the hashes. It applies to snapshots, blocks, and transactions. But worth double-checking if hashing logic is consistent for all the stored things.

sql/snapshot.sql
```diff
@@ -0,0 +1,546 @@
...
+-- DROP TABLE dag_allow_spends;
+
+CREATE TABLE dag_allow_spends (
+	last_valid_epoch_progress int8 NOT NULL,
+	approvers _varchar NOT NULL,
+	CONSTRAINT dag_allow_spends_pk PRIMARY KEY (hash),
+	CONSTRAINT destination_addr_not_null CHECK ((destination_addr IS NOT NULL)),
+	CONSTRAINT source_addr_not_null CHECK ((source_addr IS NOT NULL))
+)
+INHERITS (public.abstract_transactions);
```
kpudlik, line 169: `dag_allow_spends` inherits the `created_at` and `updated_at` but won't inherit the `DEFAULT now()` afaik. Double-check please.

sql/snapshot.sql
```diff
@@ -0,0 +1,546 @@
...
+
+-- DROP TABLE dag_reward_transactions;
+
+CREATE TABLE dag_reward_transactions (
+	global_snapshot_hash varchar NOT NULL,
+	CONSTRAINT dag_reward_transaction_pk PRIMARY KEY (hash),
+	CONSTRAINT destination_addr_not_null CHECK ((destination_addr IS NOT NULL)),
+	CONSTRAINT reward_transaction_global_snapshot_fk FOREIGN KEY (global_snapshot_hash) REFERENCES global_snapshots(hash)
+)
+INHERITS (public.abstract_transactions);
```
kpudlik: inherits the `created_at` and `updated_at` but won't inherit the `DEFAULT now()` afaik. Double-check please.

sql/snapshot.sql
```diff
@@ -0,0 +1,546 @@
...
+-- DROP TABLE dag_spend_transactions;
+
+CREATE TABLE dag_spend_transactions (
+	allow_spend varchar NULL,
+	CONSTRAINT dag_spend_transactions_pk PRIMARY KEY (hash),
+	CONSTRAINT destination_addr_not_null CHECK ((destination_addr IS NOT NULL)),
+	CONSTRAINT original_not_null CHECK ((original IS NOT NULL)),
+	CONSTRAINT dag_spend_transactions_dag_allow_spends_fk FOREIGN KEY (allow_spend) REFERENCES dag_allow_spends(hash)
+)
+INHERITS (public.abstract_transactions);
```
kpudlik, line 276: inherits the `created_at` and `updated_at` but won't inherit the `DEFAULT now()` afaik. Double-check please.

sql/snapshot.sql
```diff
@@ -0,0 +1,546 @@
...
+-- DROP TABLE dag_token_locks;
+
+CREATE TABLE dag_token_locks (
+	ordinal int8 NOT NULL,
+	unlock_epoch int8 NOT NULL,
+	CONSTRAINT dag_token_locks_pk PRIMARY KEY (hash),
+	CONSTRAINT dag_token_locks_unique UNIQUE (hash, ordinal),
+	CONSTRAINT source_addr_not_null CHECK ((source_addr IS NOT NULL))
+)
+INHERITS (public.abstract_transactions);
```
kpudlik, line 322: inherits the `created_at` and `updated_at` but won't inherit the `DEFAULT now()` afaik. Double-check please.

sql/snapshot.sql
```diff
@@ -0,0 +1,546 @@
...
+-- Drop table
+
+-- DROP TABLE dag_blocks;
+
+CREATE TABLE dag_blocks (
+	snapshot_hash varchar NOT NULL,
+	CONSTRAINT dag_block_pk PRIMARY KEY (hash),
+	CONSTRAINT dag_block_global_snapshot_fk FOREIGN KEY (hash) REFERENCES global_snapshots(hash)
+)
+INHERITS (public.abstract_blocks);
```
kpudlik, line 222: inherits the `created_at` and `updated_at` but won't inherit the `DEFAULT now()` afaik. Double-check please.

sql/snapshot.sql
```diff
@@ -0,0 +1,546 @@
...
+
+-- DROP TABLE metagraph_blocks;
+
+CREATE TABLE metagraph_blocks (
+	metagraph_id varchar NOT NULL,
+	snapshot_hash varchar NOT NULL,
+	CONSTRAINT metagraph_block_pk PRIMARY KEY (metagraph_id, hash),
+	CONSTRAINT metagraph_block_metagraph_snapshot_fk FOREIGN KEY (metagraph_id,snapshot_hash) REFERENCES metagraph_snapshots(metagraph_id,hash)
+)
+INHERITS (public.abstract_blocks);
```
kpudlik, line 685: inherits the `created_at` and `updated_at` but won't inherit the `DEFAULT now()` afaik. Double-check please.

sql/snapshot.sql
```diff
@@ -0,0 +1,546 @@
...
+-- DROP TABLE fee_transactions;
+
+CREATE TABLE fee_transactions (
+	metagraph_id varchar NOT NULL,
+	metagraph_snapshot_hash varchar NOT NULL,
+	parent jsonb NOT NULL,
+	CONSTRAINT fee_transaction_pk PRIMARY KEY (hash, metagraph_id),
+	CONSTRAINT fee_transaction_metagraph_snapshot_fk FOREIGN KEY (metagraph_id,metagraph_snapshot_hash) REFERENCES metagraph_snapshots(metagraph_id,hash)
+)
+INHERITS (public.abstract_transactions);
```
kpudlik, line 630: inherits the `created_at` and `updated_at` but won't inherit the `DEFAULT now()` afaik. Double-check please.

sql/snapshot.sql
```diff
@@ -0,0 +1,546 @@
...
+-- public.fee_transactions definition
+
+-- Drop table
+
+-- DROP TABLE fee_transactions;
+
+CREATE TABLE fee_transactions (
+	metagraph_id varchar NOT NULL,
+	metagraph_snapshot_hash varchar NOT NULL,
+	parent jsonb NOT NULL,
```
kpudlik: `parent` should have `hash` and `ordinal`. So why jsonb for data that is well structured and won't change dynamically? 🤔

sql/snapshot.sql
```diff
@@ -0,0 +1,546 @@
...
+-- DROP TABLE abstract_transactions;
+
+CREATE TABLE abstract_transactions (
+	hash varchar NOT NULL,
+	source_addr varchar NULL,
+	destination_addr varchar NULL,
+	amount int8 NOT NULL,
+	fee int8 NULL,
+	salt int8 NULL,
+	original jsonb NULL,
```
kpudlik: This field existed only for `transactions` (double check if we still need it, @marcinwadon?).
It should not exist for `fee_transactions` but it exists because of inheritance.

sql/snapshot.sql
```diff
@@ -0,0 +1,546 @@
+-- public.abstract_blocks definition
+
+-- Drop table
+
+-- DROP TABLE abstract_blocks;
+
+CREATE TABLE abstract_blocks (
+	hash varchar NOT NULL,
+	height int8 NOT NULL,
+	parent jsonb NULL,
```
kpudlik: Previously parent had `hash` and `height`. So why jsonb for data that is well structured and won't change dynamically? 🤔

kpudlik (changes requested):

sql/snapshot.sql
AlexBrandes: We looked at that option in a previous review and discovered it's actually not any more efficient to use varchar(64) rather than varchar in postgres.

sql/snapshot.sql
```diff
@@ -0,0 +1,546 @@
...
+);
+
+
+-- public.abstract_transactions definition
+
+-- Drop table
+
+-- DROP TABLE abstract_transactions;
+
+CREATE TABLE abstract_transactions (
```
AlexBrandes, line 24: There are a lot of tables inheriting from this abstract_transactions table but they are inheriting more columns than they should have: `salt`, `original`, `parent_ordinal`, etc. Is the idea that the inheriting tables would use null values for columns that don't apply?

sql/snapshot.sql
```diff
@@ -0,0 +1,546 @@
...
+-- public.dag_blocks definition
+
+-- Drop table
+
+-- DROP TABLE dag_blocks;
+
+CREATE TABLE dag_blocks (
+	snapshot_hash varchar NOT NULL,
+	CONSTRAINT dag_block_pk PRIMARY KEY (hash),
+	CONSTRAINT dag_block_global_snapshot_fk FOREIGN KEY (hash) REFERENCES global_snapshots(hash)
```
AlexBrandes: Should this be `FOREIGN KEY (snapshot_hash)...`?

sql/snapshot.sql
AlexBrandes: Should it just be `parent_hash` with a FK?

sql/snapshot.sql
```diff
@@ -0,0 +1,546 @@
...
+
+-- public.abstract_transactions definition
+
+-- Drop table
+
+-- DROP TABLE abstract_transactions;
+
+CREATE TABLE abstract_transactions (
+	hash varchar NOT NULL,
+	source_addr varchar NULL,
```
AlexBrandes: Are we going to have an `addresses` table? If so, these should FK to it. I would also suggest using a consistent form like `source_address` and `destination_address` if it's referencing the address column on the addresses table.

sql/snapshot.sql
```diff
@@ -0,0 +1,546 @@
...
+
+-- public.dag_allow_spends definition
+
+-- Drop table
+
+-- DROP TABLE dag_allow_spends;
+
+CREATE TABLE dag_allow_spends (
+	last_valid_epoch_progress int8 NOT NULL,
+	approvers _varchar NOT NULL,
```
AlexBrandes:
```suggestion
	approvers varchar NOT NULL,
```

sql/snapshot.sql
```diff
@@ -0,0 +1,546 @@
...
+
+
+-- public.dag_reward_transactions definition
+
+-- Drop table
+
+-- DROP TABLE dag_reward_transactions;
+
+CREATE TABLE dag_reward_transactions (
+	global_snapshot_hash varchar NOT NULL,
```
AlexBrandes, line 244: We should be consistent with the naming across tables. If we're going to do `global_snapshot_hash` here then all other hashes that are FKs to global snapshots should also be `global_snapshot_hash` rather than `snapshot_hash`. For example, `dag_blocks` above and others. 

My preference is to be explicit like this so it's easy to see what a column FKs to at a glance (global_snapshot_hash, metagraph_snapshot_hash, etc.) but being consistent is most important.

sql/snapshot.sql
```diff
@@ -0,0 +1,546 @@
...
+	CONSTRAINT dag_reward_transaction_pk PRIMARY KEY (hash),
```
AlexBrandes: Reward transactions don't actually have a hash. For example, see rewards array here: 
http://alnilam-node.constellationnetwork.io:9000/global-snapshots/3734676

sql/snapshot.sql
```diff
@@ -0,0 +1,546 @@
...
+
+-- Drop table
+
+-- DROP TABLE dag_spend_transactions;
+
+CREATE TABLE dag_spend_transactions (
+	allow_spend varchar NULL,
+	CONSTRAINT dag_spend_transactions_pk PRIMARY KEY (hash),
+	CONSTRAINT destination_addr_not_null CHECK ((destination_addr IS NOT NULL)),
+	CONSTRAINT original_not_null CHECK ((original IS NOT NULL)),
```
AlexBrandes: We don't want to save this for anything except dag transactions.

sql/snapshot.sql
```diff
@@ -0,0 +1,546 @@
...
+
+
+-- public.dag_spend_transactions definition
+
+-- Drop table
+
+-- DROP TABLE dag_spend_transactions;
+
+CREATE TABLE dag_spend_transactions (
+	allow_spend varchar NULL,
```
AlexBrandes:
```suggestion
	allow_spend_hash varchar NULL,
```

sql/snapshot.sql
```diff
@@ -0,0 +1,546 @@
...
+
+-- Drop table
+
+-- DROP TABLE dag_token_locks;
+
+CREATE TABLE dag_token_locks (
+	ordinal int8 NOT NULL,
+	unlock_epoch int8 NOT NULL,
+	CONSTRAINT dag_token_locks_pk PRIMARY KEY (hash),
+	CONSTRAINT dag_token_locks_unique UNIQUE (hash, ordinal),
```
AlexBrandes, line 318: Isn't this redundant because hash will always be unique?

sql/snapshot.sql
```diff
@@ -0,0 +1,546 @@
...
+
+CREATE TABLE dag_token_unlocks (
+	lock_reference_ordinal int8 NOT NULL,
+	lock_reference_hash varchar NOT NULL,
+	amount int8 NOT NULL,
+	address varchar NOT NULL,
+	created_at timestamp DEFAULT now() NOT NULL,
+	updated_at timestamp DEFAULT now() NOT NULL,
+	CONSTRAINT dag_token_unlocks_pk PRIMARY KEY (lock_reference_ordinal, lock_reference_hash),
+	CONSTRAINT dag_token_unlocks_token_locks_fk FOREIGN KEY (lock_reference_hash,lock_reference_ordinal) REFERENCES dag_token_locks(hash,ordinal)
```
AlexBrandes: Is there any need to use a compound FK here when `hash` is already the PK and unique?

sql/snapshot.sql
```diff
@@ -0,0 +1,546 @@
...
+-- Drop table
+
+-- DROP TABLE metagraph_spend_transactions;
+
+CREATE TABLE metagraph_spend_transactions (
+	metagraph_id varchar NOT NULL,
+	allow_spend varchar NULL,
+	CONSTRAINT destination_addr_not_null CHECK ((destination_addr IS NOT NULL)),
+	CONSTRAINT metagraph_spend_transactions_pk PRIMARY KEY (hash),
+	CONSTRAINT original_not_null CHECK ((original IS NOT NULL)),
```
AlexBrandes: Only for dag transactions

sql/snapshot.sql
```diff
@@ -0,0 +1,546 @@
...
+
+-- public.metagraph_spend_transactions definition
+
+-- Drop table
+
+-- DROP TABLE metagraph_spend_transactions;
+
+CREATE TABLE metagraph_spend_transactions (
+	metagraph_id varchar NOT NULL,
+	allow_spend varchar NULL,
```
AlexBrandes:
```suggestion
	metagraph_allow_spend_hash varchar NULL,
```

sql/snapshot.sql
```diff
@@ -0,0 +1,546 @@
...
+
+
+-- public.metagraph_token_locks definition
+
+-- Drop table
+
+-- DROP TABLE metagraph_token_locks;
+
+CREATE TABLE metagraph_token_locks (
+	ordinal int8 NOT NULL,
```
AlexBrandes: Is this global_ordinal or metagraph_ordinal?

sql/snapshot.sql
```diff
@@ -0,0 +1,546 @@
...
+	metagraph_id varchar NOT NULL,
+	unlock_epoch int8 NOT NULL,
+	CONSTRAINT metagraph_token_locks_pk PRIMARY KEY (hash),
+	CONSTRAINT metagraph_token_locks_unique UNIQUE (hash, ordinal),
```
AlexBrandes: I think `hash` is already unique here too.

sql/snapshot.sql
```diff
@@ -0,0 +1,546 @@
...
+
+-- DROP TABLE proofs;
+
+CREATE TABLE proofs (
+	id varchar NOT NULL,
+	signature varchar NOT NULL,
+	snapshot_hash varchar NOT NULL,
+	created_at timestamp DEFAULT now() NOT NULL,
+	updated_at timestamp DEFAULT now() NOT NULL,
+	CONSTRAINT proof_pk PRIMARY KEY (id, signature, snapshot_hash),
```
AlexBrandes: I think just snapshot hash and id should be unique. A node can't sign a snapshot multiple times (or shouldn't be able to). 
```suggestion
	CONSTRAINT proof_pk PRIMARY KEY (snapshot_hash, id),
```

sql/snapshot.sql
```diff
@@ -0,0 +1,546 @@
...
+    public.metagraph_token_unlocks for each row execute function update_updated_at_column();
+
+
+-- public.proofs definition
+
+-- Drop table
+
+-- DROP TABLE proofs;
+
+CREATE TABLE proofs (
```
AlexBrandes: How about `snapshot_proofs` since this is specifically signatures for snapshots, not some other resource.

sql/snapshot.sql
AlexBrandes: Yeah, agree. It should be parent_hash and parent_ordinal as varchars.

sql/snapshot.sql
```diff
@@ -0,0 +1,546 @@
...
+
+-- Drop table
+
+-- DROP TABLE fee_transactions;
+
+CREATE TABLE fee_transactions (
+	metagraph_id varchar NOT NULL,
+	metagraph_snapshot_hash varchar NOT NULL,
+	parent jsonb NOT NULL,
+	CONSTRAINT fee_transaction_pk PRIMARY KEY (hash, metagraph_id),
```
AlexBrandes:
```suggestion
	CONSTRAINT fee_transaction_pk PRIMARY KEY (metagraph_id, hash),
```

sql/snapshot.sql
```diff
@@ -0,0 +1,546 @@
...
+
+-- public.metagraph_balance_changes definition
+
+-- Drop table
+
+-- DROP TABLE metagraph_balance_changes;
+
+CREATE TABLE metagraph_balance_changes (
+	metagraph_id varchar NOT NULL,
+	metagraph_hash varchar NOT NULL,
```
AlexBrandes:
```suggestion
	metagraph_snapshot_hash varchar NOT NULL,
```

sql/snapshot.sql
```diff
@@ -0,0 +1,546 @@
...
+
+-- public.metagraph_blocks definition
+
+-- Drop table
+
+-- DROP TABLE metagraph_blocks;
+
+CREATE TABLE metagraph_blocks (
+	metagraph_id varchar NOT NULL,
+	snapshot_hash varchar NOT NULL,
```
AlexBrandes:
```suggestion
	metagraph_snapshot_hash varchar NOT NULL,
```

sql/snapshot.sql
```diff
@@ -0,0 +1,546 @@
...
+-- public.metagraph_reward_transactions definition
+
+-- Drop table
+
+-- DROP TABLE metagraph_reward_transactions;
+
+CREATE TABLE metagraph_reward_transactions (
+	metagraph_id varchar NOT NULL,
+	metagraph_snapshot_hash varchar NOT NULL,
+	CONSTRAINT metagraph_reward_transaction_pk PRIMARY KEY (hash),
```
AlexBrandes: These don't have a `hash`

sql/snapshot.sql
```diff
@@ -0,0 +1,546 @@
...
+-- Drop table
+
+-- DROP TABLE metagraph_transactions;
+
+CREATE TABLE metagraph_transactions (
+	metagraph_id varchar NOT NULL,
+	block_hash varchar NOT NULL,
+	CONSTRAINT destination_addr_not_null CHECK ((destination_addr IS NOT NULL)),
+	CONSTRAINT metagraph_transaction_pk PRIMARY KEY (hash),
+	CONSTRAINT original_not_null CHECK ((original IS NOT NULL)),
```
AlexBrandes: We don't want to save an original transaction for anything except dag transactions (for backwards compatibility). Otherwise it's just non-normalized data.

sql/snapshot.sql
gclaramunt: approvers is an array in AllowSpend, if we're just storing, the type  array of strings should be fine. But maybe is worthwhile to add a relation table and do a FK with the address table.

sql/snapshot.sql
```diff
@@ -0,0 +1,707 @@
...
+
+
+-- public.addresses definition
+
+-- Drop table
+
+-- DROP TABLE addresses;
+
+CREATE TABLE addresses (
+	hash varchar NOT NULL,
```
AlexBrandes, line 25: Let's call this `address` - addresses are a bit different from hashes.

sql/snapshot.sql
```diff
@@ -0,0 +1,707 @@
...
+-- Drop table
+
+-- DROP TABLE dag_reward_transactions;
+
+CREATE TABLE dag_reward_transactions (
+	global_snapshot_hash varchar NOT NULL,
+	CONSTRAINT dag_reward_transaction_pk PRIMARY KEY (hash),
+	CONSTRAINT reward_transaction_global_snapshot_fk FOREIGN KEY (global_snapshot_hash) REFERENCES global_snapshots(hash)
+)
+INHERITS (public.abstract_transactions);
```
AlexBrandes: I think it might simplify things if we don't try to have reward transactions inherit from the abstract class. It's an odd one with no source address or hash.

PR #74, opened 2025-02-26
feat: adding last n incremental snapshot directory

marcus-girolneto:


marcus-girolneto:
We will not need this anymore, since we solve the sync issue

PR #77, opened 2025-03-16
feat: adding delegated staking to GL0

marcus-girolneto:


src/main/scala/org/constellation/snapshotstreaming/TessellationServices.scala
```diff
@@ -32,6 +32,7 @@ object TessellationServices {
     for {
       _ <- Async[F].unit
       txHasher = Hasher.forKryo
+      withdrawalTimeLimit = NonNegLong.unsafeFrom(136080)
```
gclaramunt: move to config?

gclaramunt (approved):
LGTM

src/main/scala/org/constellation/snapshotstreaming/TessellationServices.scala
marcus-girolneto: 👍🏼

PR #82, opened 2025-04-21
add create and withdraw delegated staking with balances

gclaramunt:
Add support for staking in snapshot streaming.

### DIFF
--- a/sql/delegated_staking.sql
+++ b/sql/delegated_staking.sql
@@ -1,0 +1,65 @@
+
+CREATE TABLE delegate_stake_create_events (
+    hash varchar PRIMARY KEY,
+    ordinal int8 NOT NULL,
+    source_addr varchar NOT NULL REFERENCES addresses(address) ON DELETE CASCADE,
+    node_id varchar NOT NULL,
+    amount int8 NOT NULL,
+    fee int8 NOT NULL DEFAULT 0,
+    lock_reference_hash varchar NOT NULL REFERENCES dag_token_locks(hash) ON DELETE CASCADE,
+    parent_hash varchar NOT NULL,
+    global_snapshot_hash varchar NOT NULL REFERENCES global_snapshots(hash) ON DELETE CASCADE,
+    is_update boolean NOT NULL,
+    created_at timestamp DEFAULT now() NOT NULL,
+    updated_at timestamp DEFAULT now() NOT NULL
+);
+CREATE INDEX delegate_stake_create_events_source_addr_idx ON public.delegate_stake_create_events USING btree (source_addr);
+CREATE INDEX delegate_stake_create_events_lock_reference_hash_idx ON public.delegate_stake_create_events USING btree (lock_reference_hash);
+CREATE INDEX delegate_stake_create_events_global_snapshot_hash_idx ON public.delegate_stake_create_events USING btree (global_snapshot_hash);
+
+
+
+CREATE TABLE delegate_stake_withdraw_events (
+    hash varchar PRIMARY KEY,
+    source_addr varchar NOT NULL REFERENCES addresses(address) ON DELETE CASCADE,
+    stake_create_hash varchar NOT NULL REFERENCES delegate_stake_create_events(hash) ON DELETE CASCADE,
+    global_snapshot_hash varchar NOT NULL REFERENCES global_snapshots(hash) ON DELETE CASCADE,
+    created_at timestamp DEFAULT now() NOT NULL,
+    updated_at timestamp DEFAULT now() NOT NULL
+);
+CREATE INDEX delegate_stake_withdraw_events_source_addr_idx ON public.delegate_stake_withdraw_events USING btree (source_addr);
+CREATE INDEX delegate_stake_withdraw_events_stake_create_hash_idx ON public.delegate_stake_withdraw_events USING btree (stake_create_hash);
+CREATE INDEX delegate_stake_withdraw_events_global_snapshot_hash_idx ON public.delegate_stake_withdraw_events USING btree (global_snapshot_hash);
+
+
+CREATE TABLE delegate_stake_balance_changes (
+	global_snapshot_hash varchar NOT NULL REFERENCES global_snapshots(hash) ON DELETE CASCADE,
+	global_snapshot_ordinal int8 NOT NULL REFERENCES global_snapshots(ordinal) ON DELETE CASCADE,
+	address varchar NOT NULL REFERENCES addresses(address) ON DELETE CASCADE,
+	node_id varchar NOT NULL,
+	balance int8 NOT NULL,
+	rewards int8 NOT NULL,
+	stake_create_hash varchar NOT NULL REFERENCES delegate_stake_create_events(hash) ON DELETE CASCADE,
+	stake_withdraw_hash varchar NOT NULL REFERENCES delegate_stake_withdraw_events(hash) ON DELETE CASCADE,
+	created_at timestamp DEFAULT now() NOT NULL,
+	updated_at timestamp DEFAULT now() NOT NULL
+);
+CREATE INDEX delegate_stake_balance_changes_global_snapshot_hash_idx ON public.delegate_stake_balance_changes USING btree (global_snapshot_hash);
+CREATE INDEX delegate_stake_balance_changes_global_snapshot_ordinal_idx ON public.delegate_stake_balance_changes USING btree (global_snapshot_ordinal);
+CREATE INDEX delegate_stake_balance_changes_address ON public.delegate_stake_balance_changes USING btree (address);
+CREATE INDEX delegate_stake_balance_changes_node_id ON public.delegate_stake_balance_changes USING btree (node_id);
+ALTER TABLE public.delegate_stake_balance_changes ADD CONSTRAINT delegate_stake_balance_changes_unique UNIQUE (global_snapshot_hash, address, node_id, balance, rewards);
+
+
+CREATE TABLE delegate_stake_rewards (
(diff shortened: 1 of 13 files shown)


gclaramunt:
I'm working on splitting staking balances from rewards (initially joined for easier querying)

sql/delegated_staking.sql
```diff
@@ -0,0 +1,71 @@
+--
+--We want to be able to show delegated staking creation, update, and withdrawal events in the DAG explorer so we’ll need support for them through the API.
```
AlexBrandes: Are we using migrations for these database changes? How is the full database schema represented in the codebase?

sql/delegated_staking.sql
```diff
@@ -0,0 +1,71 @@
...
+--
+--/addresses/:address/delegated-stakes
+--
+--/delegated-stake-withdrawals
+--
+--/delegated-stake-withdrawals/:hash
+--
+--/addresses/:address/delegated-stake-withdrawals
+
+CREATE TABLE delegate_stake_create_events (
```
AlexBrandes, line 2: How are update events represented here?

sql/delegated_staking.sql
```diff
@@ -0,0 +1,71 @@
...
+    hash varchar PRIMARY KEY,
+    ordinal int8 NOT NULL,
+    source_addr varchar NOT NULL REFERENCES addresses(address) ON DELETE CASCADE, --FK address
+    node_id varchar NOT NULL,
+    amount int8 NOT NULL,
+    fee int8 NOT NULL DEFAULT 0,
+    token_lock_ref varchar NOT NULL REFERENCES dag_token_locks(hash) ON DELETE CASCADE, --FK token lock
+    parent_hash varchar NOT NULL,
+    global_snapshot_hash varchar NOT NULL REFERENCES global_snapshots(hash) ON DELETE CASCADE
```
AlexBrandes: Don't we need `created_at` and `updated_at` here?

sql/delegated_staking.sql
```diff
@@ -0,0 +1,71 @@
...
+);
+CREATE INDEX delegate_stake_create_events_source_addr_idx ON public.delegate_stake_create_events USING btree (source_addr);
+CREATE INDEX delegate_stake_create_events_token_lock_ref_idx ON public.delegate_stake_create_events USING btree (token_lock_ref);
+CREATE INDEX delegate_stake_create_events_global_snapshot_hash_idx ON public.delegate_stake_create_events USING btree (global_snapshot_hash);
+
+
+
+CREATE TABLE delegate_stake_withdraw_events (
```
AlexBrandes, line 22: `created_at` and `updated_at`?

sql/delegated_staking.sql
```diff
@@ -0,0 +1,71 @@
...
+CREATE INDEX delegate_stake_withdraw_events_global_snapshot_hash_idx ON public.delegate_stake_withdraw_events USING btree (global_snapshot_hash);
+
+
+CREATE TABLE delegate_stake_balance_changes (
+	global_snapshot_hash varchar NOT NULL REFERENCES global_snapshots(hash) ON DELETE CASCADE,
+	global_snapshot_ordinal int8 NOT NULL REFERENCES global_snapshots(ordinal) ON DELETE CASCADE
+	address varchar NOT NULL REFERENCES addresses(address) ON DELETE CASCADE,
+	node_id varchar NOT NULL,
+	balance int8 NOT NULL,
+	rewards int8 NOT NULL,
```
AlexBrandes, line 41: Does `balance` represent the TokenLock amount and `rewards` represents the accumulated rewards?

sql/delegated_staking.sql
```diff
@@ -0,0 +1,71 @@
...
+	created_at timestamp DEFAULT now() NOT NULL,
+	updated_at timestamp DEFAULT now() NOT NULL
+);
+CREATE INDEX delegate_stake_balance_changes_idx ON public.delegate_stake_balance_changes USING btree (address);
```
AlexBrandes: NodeID will be the other way we'll need to query this so it would make sense to add an index for that here.

sql/delegated_staking.sql
```diff
@@ -0,0 +1,71 @@
...
+CREATE INDEX delegate_stake_balance_changes_global_snapshot_hash_idx ON public.delegate_stake_balance_changes USING btree (global_snapshot_hash);
+CREATE INDEX delegate_stake_balance_changes_global_snapshot_ordinal_idx ON public.delegate_stake_balance_changes USING btree (global_snapshot_ordinal);
+ALTER TABLE public.delegate_stake_balance_changes ADD CONSTRAINT delegate_stake_balance_changes_unique UNIQUE (address,node_id,balance,rewards);
+
+
+
+
+ALTER TABLE public.delegate_stake_balance_changes ADD global_snapshot_ordinal int8 NULL;
```
AlexBrandes: Why is it being made nullable here?

sql/delegated_staking.sql
gclaramunt: yes, accumulated reward for this particular stake (as I understand) they come in the calculated state as
```
case class DelegatedStakeRecord(
    event: Signed[UpdateDelegatedStake.Create],
    createdAt: SnapshotOrdinal,
    rewards: Balance
  )
```

sql/delegated_staking.sql
gclaramunt: I don't really differentiate it here. We can rebuild it from a query (by same token_lock_ref) <del>but probably is easier to have a nullable "previous_stake_hash" a create will have null and an update will have the hash of the previous stake </del>. (update: if we don't need to rebuild the history, just a boolean flag should be enough)

sql/delegated_staking.sql
gclaramunt: I'm keeping the full schema in https://github.com/Constellation-Labs/snapshot-streaming/blob/release/testnet/sql/snapshot.sql 

I can add flyway migration at startup and control through it. What do you think?
(we can leave mainnet out of it, just in case)

src/main/scala/org/constellation/snapshotstreaming/mapper/GlobalSnapshotMapper.scala
```diff
@@ -46,11 +62,145 @@ abstract class GlobalSnapshotMapper[F[_]: Async]
...
+    gsHash: Hash,
+    gsOrdinal: SnapshotOrdinal,
+    snapshotInfo: GlobalSnapshotInfo,
+    maybePrevSnapshotInfo: Option[GlobalSnapshotInfo]
+  )(implicit hasher: Hasher[F]) = {
+
+    val prev = maybePrevSnapshotInfo.toSeq.flatMap(s => flatten(s.activeDelegatedStakes))
+    val current = flatten(snapshotInfo.activeDelegatedStakes)
+
+    // might not be necessary if rewards are always updating the current balance ?
```
AlexBrandes: Only on timed snapshots

sql/delegated_staking.sql
```diff
@@ -0,0 +1,49 @@
+
+CREATE TABLE delegate_stake_create_events (
+    hash varchar PRIMARY KEY,
+    ordinal int8 NOT NULL,
```
AlexBrandes, line 4: Why is this `ordinal` vs `global_snapshot_hash` below? Shouldn't we be explicit that this is the global ordinal also?

sql/delegated_staking.sql
```diff
@@ -0,0 +1,49 @@
...
+CREATE INDEX delegate_stake_create_events_token_lock_ref_idx ON public.delegate_stake_create_events USING btree (token_lock_ref);
+CREATE INDEX delegate_stake_create_events_global_snapshot_hash_idx ON public.delegate_stake_create_events USING btree (global_snapshot_hash);
+
+
+
+CREATE TABLE delegate_stake_withdraw_events (
+    hash varchar PRIMARY KEY,
+    source_addr varchar NOT NULL REFERENCES addresses(address) ON DELETE CASCADE,
+    stake_ref varchar NOT NULL REFERENCES delegate_stake_create_events(hash) ON DELETE CASCADE,
+    global_snapshot_hash varchar NOT NULL REFERENCES global_snapshots(hash) ON DELETE CASCADE,
```
AlexBrandes, line 26: We should be consistent and track `global_snapshot_ordinal` here if we're going to do it on the create event.

sql/delegated_staking.sql
AlexBrandes: We'll need to set up an initial migration and then all future database changes get added as a new migration in a PR. Not sure about running it at startup - it's just to make sure that the full database schema can be replicated and rolled back if necessary.

sql/delegated_staking.sql
gclaramunt: this is actually a delegate stake ordinal, different from the snapshot ordinal, it comes from tessellation Create record:
```
@derive(eqv, show, encoder, decoder)
    case class Create(
      source: Address,
      nodeId: PeerId,
      amount: DelegatedStakeAmount,
      fee: DelegatedStakeFee = DelegatedStakeFee(NonNegLong(0L)),
      tokenLockRef: Hash,
      parent: DelegatedStakeReference = DelegatedStakeReference.empty
    ) extends UpdateDelegatedStake {
      def ordinal: DelegatedStakeOrdinal = parent.ordinal.next
    }
 ```

sql/delegated_staking.sql
gclaramunt: the withdrawal record doesn't have ordinal, so is not there. For now, I'm trying to avoid tracking snapshot_ordinal if I have the snapshot_hash, unless it's directly present in the schema

AlexBrandes (approved):
LGTM - we can sort out migrations later.

marcus-girolneto (approved):

PR #93, opened 2025-05-22
feat: add snapshot queue for downloaded snapshots

AlexBrandes:
Splits snapshot processing into two processes that pass data through a queue:
- Producer that downloads and validates snapshots then pushes them to a queue
- Consumer that pulls snapshots in sequence from the queue as they become available

### DIFF
--- a/src/main/scala/org/constellation/snapshotstreaming/SnapshotProcessor.scala
+++ b/src/main/scala/org/constellation/snapshotstreaming/SnapshotProcessor.scala
@@ -4,10 +4,17 @@ import java.time.{Instant, LocalDateTime, ZoneId}
 import cats.data.{NonEmptyList, Validated}
 import cats.effect._
 import cats.effect.implicits.clockOps
-import cats.effect.std.{Console, Random}
+import cats.effect.std.{Console, Queue, Random}
 import cats.syntax.all._
-import cats.{Applicative, Parallel}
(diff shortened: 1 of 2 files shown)


src/main/scala/org/constellation/snapshotstreaming/SnapshotProcessor.scala
```diff
@@ -202,111 +217,142 @@ object SnapshotProcessor {
...
+          .evalTap { snapshot =>
+            queue.size.flatMap { size =>
+              logger.info(
+                s"Consumer: Starting to process snapshot ${getSnapshotReference(snapshot.snapshot)}. Queue size: $size"
+              )
+            }
+          }
+          .evalMap { snapshot =>
+            val hasher = HasherSelector[F].getForOrdinal(snapshot.snapshot.ordinal)
+            store(snapshot, hasher)
```
ryle-goehausen: Is this store lazily evaluated ? Need to load this in IDE, but this call isn't pipelined with `>>`

ryle-goehausen: Yeah it's an F return, so need >>

AlexBrandes: Good catch - fixed

marcus-girolneto (approved):
Looks good to me, great idea btw

PR #98, opened 2025-05-23
parallelize db inserts

gclaramunt:
now the store(snapshot) can be done in parallel, parEvalMap preserves ordering, so storing the checkpoint later is safe.
(the parEvalMap approach is used succesfully in the reindexer, so I don't expect problems here)


AlexBrandes (changes requested):
I don’t think this is a good idea because it breaks the sequential insertion of snapshots. We’ll end up with FK conflicts and other potential issues. 

The current configuration starts processing each snapshot from the queue as soon as it’s added and the insert process is faster than snapshot validation so the queue essentially never backs up (that we’ve seen so far anyways). This adds risk without much benefit imo.

PR #100, opened 2025-05-24
Remove duplicate processing

gclaramunt:


src/main/scala/org/constellation/snapshotstreaming/SnapshotProcessor.scala
```diff
@@ -333,19 +322,14 @@ object SnapshotProcessor {
...
-              logger.info(s"Consumer: Storing snapshot ${getSnapshotReference(snapshot.snapshot)}") >>
-                FileBasedLastGlobalIncrementalSnapshotStorage
-                  .saveSnapshotWithStateJson(
-                    Path(s"snapshotWithState.$snapshotOrdinal.json.gz"),
-                    snapshotWithState,
-                    Flags.Write
-                  )
-                  .whenA(snapshotOrdinal % configuration.checkpointEvery == 0)
-            }
+            logger.info(s"Producer: Processing snapshot ${getSnapshotReference(snapshot.snapshot)}") >>
```
AlexBrandes:
```suggestion
            logger.info(s"Consumer: Processing snapshot ${getSnapshotReference(snapshot.snapshot)}") >>
```

AlexBrandes (commented):
This seems important to fix and would explain a lot of the strange behavior we've seen. Looks good to me but I'll let someone else review and give an approval.

marcus-girolneto (approved):

PR #107, opened 2025-06-03
PROT-1251 update delegated stake withdraw

gclaramunt:
Emit the list of completed withdrawals hashes from the mapper and run an update with those setting the completed flag to true


AlexBrandes (approved):
LGTM

PR #114, opened 2025-12-15
feat: capture delegate stake increase and lock transfer

gclaramunt:


ryle-ai (commented):
Decent-sized PR -- three concerns worth flagging:

**1. The `ON CONFLICT DO UPDATE` in `insertDelegatedStakeCreateCommand` is a meaningful behavior change**

In `SnapshotDAO.scala`, the delegate_stake_create_events insert went from `ON CONFLICT (hash) DO NOTHING` to `ON CONFLICT (hash) DO UPDATE SET current_lock_reference_hash = EXCLUDED.current_lock_reference_hash, current_amount = EXCLUDED.current_amount, transfer_from_hash = EXCLUDED.transfer_from_hash, updated_at = now()`. This means if the same stake hash appears again with different `currentTokenLockRef`/`currentAmount` values (i.e., the stake was increased), we update the existing row rather than silently ignoring it.

This makes sense for the "stake increase" use case -- the `DelegatedStakeRecord` keeps the same `event` (and thus the same hash) but gets new `currentTokenLockRef` and `currentAmount` fields. But I want to make sure we're intentional about also updating `transfer_from_hash` on conflict -- is there a scenario where the `transferFrom` value changes between re-inserts and we'd be overwriting a meaningful original value? I think this is probably fine since `mapDelegatedStakingCreates` in `GlobalSnapshotMapper.scala:233` computes `fromHash` fresh each time based on the previous snapshot's active stakes, but worth double-checking.

**2. The `mapTokenLock` visibility change from `private` to `public`**

In `GlobalSnapshotMapper.scala`, `mapTokenLock` went from `private def` to `def`. The only new call site I can see is in the test suite (`GlobalSnapshotMapperSuite`). If this is purely for testability, that's fine for now, but ideally we'd keep the narrower visibility and test through the public `mapTokenLocks` method instead. Not a blocker.

**3. Merge conflicts need resolution**

Like PR #119, this is currently in CONFLICTING state against develop. The base tessellation version jumps from `3.2.1-rc.2-37-b82667f1-SNAPSHOT` to `4.0.0-rc.1` which is a major version bump -- there's a lot of structural churn in this PR that comes from the tessellation upgrade (import path changes like `GlobalSnapshotStateChannelEventsProcessor` moving to `managers.global`, `SharedValidators.make` gaining an `env` parameter, `.unsorted` → `.toSeq` / `.toSortedSet`, `rewards` type changing from `Set` to `Seq`, the `L0GlobalSnapshotClient.make` signature change, etc). This makes the PR harder to review because the delegated staking feature changes are mixed in with the tessellation API migration.

I'd recommend rebasing on develop (which is already at a newer tessellation version) so we can isolate just the delegated staking capture logic. Right now it's hard to separate "things that changed because tessellation 4.0.0 changed the API" from "things that changed for the delegate stake increase feature."

The core feature logic in `GlobalSnapshotMapper.scala` -- tracking `currentTokenLockRef`/`currentAmount` on `DelegatedStakeRecord` and detecting changes in `activeHashedDelegatedStakes` by comparing both the hash AND the `currentTokenLockRef` -- looks correct and aligns with the tessellation types (`DelegatedStakeRecord` at `delegatedStake.scala:122`). Good test coverage for the new scenarios too.

🤖 *ryle-ai automated review*

sql/delegated_staking.sql
```diff
@@ -10,12 +10,15 @@ CREATE TABLE delegate_stake_create_events (
     parent_hash varchar NOT NULL,
     global_snapshot_hash varchar NOT NULL REFERENCES global_snapshots(hash) ON DELETE CASCADE,
     is_update boolean NOT NULL,
+    current_lock_reference_hash varchar,
```
ryle-ai, line 13: I think there's a cross-repo naming inconsistency here that will cause problems. This column is named `current_lock_reference_hash` in snapshot-streaming, but over in block_explorer PRs #204 and #205, the same column on the same table is called `current_token_lock_hash` -- both in the Prisma schema and in the migration SQL (`prisma/migrations/20260109/02_add_current_fields_to_delegate_stake_create_events.sql`).

Since snapshot-streaming writes the data and block_explorer reads it, whichever one runs second will either create a duplicate column or fail to find the data. You'll want to align on a single name before either PR merges. I'd lean toward `current_token_lock_hash` since it's more descriptive (it's a token lock hash, not a generic lock reference), but either way they need to match.

Same issue exists for `delegate_stake_withdraw_events` a few lines below.

sql/snapshot.sql
```diff
@@ -355,6 +355,7 @@ CREATE TABLE dag_token_locks (
 	round_id uuid NOT NULL,
 	parent_hash varchar NULL,
 	snapshot_hash varchar NOT NULL,
+	replace_token_lock_ref varchar NULL,
```
ryle-ai, line 358: Same naming issue as the delegated staking columns -- this is `replace_token_lock_ref` here, but block_explorer PR #204 adds it as `replacement_hash` (`prisma/migrations/20260109/01_add_replacement_hash_to_dag_token_locks.sql`). The Prisma schema over there also references `replacement_hash`, not `replace_token_lock_ref`.

These need to be aligned before merge. I think `replace_token_lock_ref` is actually more consistent with the tessellation naming since `TokenLock.replaceTokenLockRef` is the field name in the Scala type (`tokenLock.scala` on the `feature/increase-delegated-stake` branch).

src/main/scala/org/constellation/snapshotstreaming/mapper/GlobalSnapshotMapper.scala
```diff
@@ -196,12 +198,14 @@ abstract class GlobalSnapshotMapper[F[_]: Async] extends SnapshotMapper[F, Globa
...
-      dsr.event.tokenLockRef.value,
+      dsr.tokenLockRef.value,
       dsr.event.parent.hash.value,
-      fromHash.map(_.value)
+      fromHash.map(_.value),
+      dsr.currentTokenLockRef.map(_.value),
+      dsr.currentAmount.map(_.value)
     )
 
   def activeHashedDelegatedStakes(
```
ryle-ai, line 211: So `dsr.amount` and `dsr.tokenLockRef` already resolve to the effective values via the helper methods on `DelegatedStakeRecord` (i.e. `currentAmount.getOrElse(event.amount)` -- see tessellation `feature/increase-delegated-stake` branch, `delegatedStake.scala:130-131`). That means the main `amount` and `tokenLockHash` fields in `DelegatedStakingCreate` already contain the effective/current values.

Then `currentTokenLockHash` and `currentAmount` are set separately as `dsr.currentTokenLockRef.map(_.value)` and `dsr.currentAmount.map(_.value)` -- which are the same values again, just wrapped in `Option`. This duplication is intentional for the block_explorer to differentiate between "original" and "current" values, right? If so, I think it's fine but worth a comment here explaining the semantics: `amount/tokenLockHash` = effective (for DB primary fields), `currentTokenLockHash/currentAmount` = explicit override values (for API consumers that want to know if a stake was increased).

src/main/scala/org/constellation/snapshotstreaming/mapper/GlobalSnapshotMapper.scala
```diff
@@ -221,20 +225,22 @@ abstract class GlobalSnapshotMapper[F[_]: Async] extends SnapshotMapper[F, Globa
...
         val prevActiveTokenLocksMap = prevActiveTokenLocks.toMap
         activeDelegatedStakes.mapFilter { case (dsr, ev) => // keep only the new or the updated
           prevActiveTokenLocksMap.get(dsr.event.tokenLockRef) match {
             case None => Some((dsr, ev, None)) // new stake
-            case Some(oldStakeHash) =>
-              if (oldStakeHash == ev.hash)
-                None // active staking already included
-              else {
+            case Some((oldStakeHash, oldCurrentTokenLockRef)) =>
+              val hashChanged = oldStakeHash != ev.hash
```
ryle-ai, line 237: I think there's a subtle edge case here worth considering. The comparison uses `dsr.event.tokenLockRef` as the map key (the original token lock ref from the `Create` event), but `dsr.currentTokenLockRef` for the change detection tuple. If a stake is increased twice across consecutive snapshots (TokenRef1 -> TokenRef2 -> TokenRef3), the lookup key `dsr.event.tokenLockRef` will always be the original `TokenRef1` -- which is correct for identity. But the `currentTokenLockRefChanged` check compares `oldCurrentTokenLockRef` (which would be `Some(TokenRef2)` from the previous snapshot) against `dsr.currentTokenLockRef` (now `Some(TokenRef3)`). This should work correctly for detecting the second increase.

However, I notice the key in `prevActiveTokenLocksMap` is `dsr.event.tokenLockRef` -- the *original* token lock ref. If a different stake later reuses the same original `tokenLockRef` (which I don't think is possible given the hash chain, but worth confirming), we'd get a false match. I believe this is fine in practice since the hash on `Create` events includes the token lock ref and parent, making collisions effectively impossible.

ryle-ai (commented):
## Review: feat: capture delegate stake increase and lock transfer

The prior automated review from ryle-ai covered the major issues well (column naming mismatch, ON CONFLICT behavior change, merge conflicts). I want to add two observations that were not flagged.

### 1. `transfer_from_hash` self-reference on stake increase (GlobalSnapshotMapper.scala:233-240)

When a stake is increased (same event, new `currentTokenLockRef`), the change detection at line 233-240 correctly identifies it as an update because `currentTokenLockRefChanged` is true. However, `fromHash` is set to `Some(oldStakeHash)` -- and because the underlying `event` hasn't changed (only `currentTokenLockRef` changed), `oldStakeHash == ev.hash`. This means the `ON CONFLICT DO UPDATE` will set `transfer_from_hash = EXCLUDED.transfer_from_hash` where `EXCLUDED.transfer_from_hash` is the stake's own hash.

This creates a self-referential row: `delegate_stake_create_events.transfer_from_hash` points to its own `hash`. The Prisma schema's `delegation_self` relation supports this, but it's semantically misleading -- `transfer_from_hash` was designed to mean "this stake replaced the stake with hash X" (i.e., a different stake). A self-reference means "this stake was updated in-place" which is a different concept.

Practical impact: any block_explorer query that follows the `delegated_from` chain (e.g., to show stake history) will see a cycle for increased stakes, which could cause infinite loops if not guarded. I'd suggest either:
- Setting `transfer_from_hash` to `None` for pure increases (where only `currentTokenLockRef` changed but `ev.hash` is the same), or
- Adding a comment documenting that self-references indicate in-place increases vs. `None` for new stakes vs. a different hash for true transfers.

### 2. Deployment ordering dependency

This PR adds `current_lock_reference_hash`, `current_amount` (to `delegate_stake_create_events` and `delegate_stake_withdraw_events`), and `replace_token_lock_ref` (to `dag_token_locks` and `metagraph_token_locks`) to the SQL INSERT statements. The corresponding `ALTER TABLE` / migration to add these columns to the database must be deployed **before** this snapshot-streaming code goes live, or every INSERT will fail on the missing columns.

Since the migrations live in block_explorer (PRs #204/#205 per the prior review), both of those PRs need to be merged and the migration applied before this PR's code is deployed. Worth documenting this ordering in a deployment runbook or release ticket to avoid a production outage.

ryle-ai (changes requested):
### Missing `incrementalDelegatedStakingStartingOrdinal` parameter — compilation error

The tessellation 4.0.0-rc.1 API added `incrementalDelegatedStakingStartingOrdinal: SnapshotOrdinal` as a required parameter in two places:

1. **`SharedConfig` constructor** (tessellation `types.scala:89`) — requires this field between `lastLegacyStateProofOrdinal` and `addresses`
2. **`GlobalSnapshotContextFunctions.make`** (tessellation `GlobalSnapshotContextFunctions.scala:50`) — requires it as the 7th parameter

Neither is supplied in this PR. `Configuration.nodeSharedConfig` skips from `c.lastLegacyStateProofOrdinal` straight to `c.addresses`, and `TessellationServices.make` only passes 6 arguments to `GlobalSnapshotContextFunctions.make`. This will not compile against tessellation 4.0.0-rc.1.

The fix is to add `c.incrementalDelegatedStakingStartingOrdinal` in `Configuration.nodeSharedConfig` and `configuration.fieldsAddedOrdinals.incrementalDelegatedStakingStartingOrdinal.getOrElse(env, SnapshotOrdinal.MinValue)` (or equivalent from `SharedConfigReader`) in the `GlobalSnapshotContextFunctions.make` call.

src/main/scala/org/constellation/snapshotstreaming/mapper/GlobalSnapshotMapper.scala
```diff
@@ -196,12 +198,14 @@ abstract class GlobalSnapshotMapper[F[_]: Async] extends SnapshotMapper[F, Globa
       dsr.createdAt.value,
       dsr.event.source.value,
       dsr.event.nodeId.value.value,
-      dsr.event.amount.value,
+      dsr.amount.value,
```
ryle-ai, line 201: Switching from `dsr.event.amount.value` to `dsr.amount.value` changes what the `amount` column stores. `dsr.amount` resolves to `currentAmount.getOrElse(event.amount)` — the effective/current amount, not the original stake amount.

This is fine during normal operation because the first INSERT captures the original values and subsequent `ON CONFLICT DO UPDATE` only touches `current_*` columns. But if snapshot-streaming reprocesses from a clean database and encounters a snapshot where a stake has already been increased, the INSERT will store the *increased* amount as the base `amount` — the original value is lost.

Same issue applies to `dsr.tokenLockRef.value` on line 205 (`lock_reference_hash` column).

If you want the base columns to always reflect original values, keep `dsr.event.amount.value` and `dsr.event.tokenLockRef.value` here, and let `currentTokenLockHash`/`currentAmount` carry the overrides. If you intentionally want effective values in the base columns, add a comment explaining that — it's a non-obvious semantic change from the previous behavior.

src/main/scala/org/constellation/snapshotstreaming/db/SnapshotDAO.scala
```diff
@@ -274,11 +283,13 @@ object SnapshotDAO {
...
-      ) VALUES ($varchar, $varchar, $varchar, $varchar, $int8, $int8, $bool)
+        is_completed,
+        current_lock_reference_hash,
+        current_amount
+      ) VALUES ($varchar, $varchar, $varchar, $varchar, $int8, $int8, $bool, ${varchar.opt}, ${int8.opt})
       ON CONFLICT (hash) DO NOTHING;
     """.command.contramap { tx: DelegatedStakingWithdraw =>
-      (tx.hash, tx.sourceAddress, tx.stakeCreateHash, tx.snapshotHash, tx.createdAtEpoch, tx.unlockEpoch, tx.completed)
+      (tx.hash, tx.sourceAddress, tx.stakeCreateHash, tx.snapshotHash, tx.createdAtEpoch, tx.unlockEpoch, tx.completed, tx.currentTokenLockHash, tx.currentAmount)
     }
```
ryle-ai, line 294: The withdrawal INSERT keeps `ON CONFLICT (hash) DO NOTHING`, but the create INSERT was changed to `ON CONFLICT DO UPDATE` to track evolving `currentTokenLockRef`/`currentAmount`. This creates an asymmetry: if a stake is increased while a withdrawal is pending, the create event row gets updated with the new current values, but the withdrawal row won't.

If `PendingDelegatedStakeWithdrawal.currentTokenLockRef` can change between snapshots (e.g., the underlying stake is increased after the withdrawal was initiated), this INSERT will silently drop the update. Consider whether this should also use `ON CONFLICT DO UPDATE SET current_lock_reference_hash = EXCLUDED.current_lock_reference_hash, current_amount = EXCLUDED.current_amount, updated_at = now()` for consistency.

ryle-ai (commented):
### Deep review: additional observations

The prior reviews caught the major issues (missing `incrementalDelegatedStakingStartingOrdinal`, column naming mismatches, self-referential `transfer_from_hash`). Two additional concerns below.

PR #115, opened 2025-12-17
feat: add support for increase delegated stake feature

effects-ai:


project/Dependencies.scala
```diff
@@ -20,7 +20,7 @@ object Dependencies {
     val logstash = "7.2"
     val organizeImports = "0.6.0"
     val refined = "0.10.1"
-    val tessellation = "3.2.1-rc.2-37-b82667f1-SNAPSHOT"
+    val tessellation = "99.99.99-SNAPSHOT"
```
gclaramunt: can we use a tag? That way I don't need to guess the actual version :)

PR #116, opened 2026-01-12
chore: adding automated deploy pipeline

marcus-girolneto:


project/Dependencies.scala
```diff
@@ -20,7 +20,7 @@ object Dependencies {
     val logstash = "7.2"
     val organizeImports = "0.6.0"
     val refined = "0.10.1"
-    val tessellation = "3.2.1-rc.2-37-b82667f1-SNAPSHOT"
+    val tessellation = "3.2.1-rc.2-136-0b342c0a-SNAPSHOT"
```
gclaramunt, line 23: can we use a tag so we can point to a specific version?

gclaramunt (approved):

gclaramunt:
lgtm

project/Dependencies.scala
marcus-girolneto: hmm good idea let me work in another PR

PR #121, opened 2026-04-23
feat: store state and combined snapshot+state in S3 with configurable retention

marcus-girolneto:
## Summary

Snapshot-streaming currently uploads only the raw incremental snapshot to S3. That is not sufficient to support node rollback: both the `GlobalSnapshotInfo` (state) and the combined snapshot+state bundle are required to validate and restore. Two production incidents — nodes missing snapshots, and the streaming DB getting stuck on inserts — both benefit from having the last N states reliably retrievable from S3.

- Uploads **state** (`GlobalSnapshotInfo`) and **combined** (`SnapshotWithState`) to S3 alongside the existing snapshot upload. Each upload is independently toggled.
- Keeps the last `retentionCount` of each (default `1000`) via exact-prefix pruning on the expired ordinal — O(1) per upload, no full bucket scan.
- Key scheme `states/{ordinal}-{hash}` and `combined/{ordinal}-{hash}`: ordinal enables cheap retention, hash prevents fork collisions. S3 object user-metadata carries both `ordinal` and `hash` for tooling.
- Payloads are gzipped JSON (state is ~9.5 MB uncompressed).
- `downloadState(ordinal, hash)` / `downloadCombined(ordinal, hash)` exposed on `S3DAO` for future rollback verification callers.

## Config

New fields under `snapshotStreaming.s3` (all default to off / safe values):

```hocon
s3 {
  uploadEnabled = false          # existing — raw snapshot
  uploadStateEnabled = false     # new — GlobalSnapshotInfo
  uploadCombinedEnabled = false  # new — SnapshotWithState (rollback bundle)
  retentionCount = 1000          # new — lastN kept for state + combined
}
```

## Files changed

- `src/main/scala/org/constellation/snapshotstreaming/Configuration.scala`
- `src/main/scala/org/constellation/snapshotstreaming/s3/S3DAO.scala`
- `src/main/scala/org/constellation/snapshotstreaming/SnapshotProcessor.scala`
- `src/main/resources/reference.conf`

## Out of scope / follow-ups

- **Rollback verification call site.** `downloadState` / `downloadCombined` are exposed but no consumer wires them in. The ticket's "ensure via S3 download that state/ordinal/hash match" is presumably handled by the network-monitoring service that triggers the rollback.
- **Backfill of pre-existing ordinals.** Retention only kicks in on new uploads, so flipping the flags on a running stream at ordinal N populates S3 forward only; nothing below `N - retentionCount` is ever written. Matches "store lastN going forward." Happy to add a one-shot backfill if needed.

## Test plan

- [ ] CI passes (develop pulls a snapshot tessellation artifact that is not locally cached; CI resolves from the internal repo)
- [ ] Deploy to a non-production env with `uploadStateEnabled = true` and `uploadCombinedEnabled = true`, `retentionCount = 100` for quick verification
- [ ] Confirm `states/{ordinal}-{hash}` and `combined/{ordinal}-{hash}` objects appear with correct user-metadata
- [ ] After streaming >100 snapshots, confirm the oldest ordinals are pruned from both prefixes
- [ ] Download a combined object, gunzip, confirm JSON round-trips into `SnapshotWithState` and that ordinal/hash match the object key

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -25,6 +25,9 @@ snapshotStreaming {
   opensearch {}
   s3 {
     uploadEnabled = false
+    uploadStateEnabled = false
+    uploadCombinedEnabled = false
+    retentionCount = 1000
   }
 }
 
--- a/src/main/scala/org/constellation/snapshotstreaming/Configuration.scala
+++ b/src/main/scala/org/constellation/snapshotstreaming/Configuration.scala
@@ -35,7 +35,16 @@ final case class DbConfig(
 )
 
 final case class S3ApiConfig(endpoint: Option[String], region: Option[String], pathStyleEnabled: Option[Boolean])
-final case class S3Config(bucketRegion: String, bucketName: String, bucketDir: String, api: S3ApiConfig, uploadEnabled: Boolean)
+final case class S3Config(
+  bucketRegion: String,
+  bucketName: String,
+  bucketDir: String,
+  api: S3ApiConfig,
+  uploadEnabled: Boolean,
+  uploadStateEnabled: Boolean,
+  uploadCombinedEnabled: Boolean,
+  retentionCount: Int
+)
 
 final case class OpenSearchConfig(uri: Uri, bulkSize: Int, indexes: IndexesConfig)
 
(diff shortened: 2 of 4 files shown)


gclaramunt:
LGTM

PR #123, opened 2026-07-13
fix: validate post-sub-trie-roots state proofs, bump tessellation to 4.1.0-rc.4

scasplte2:
## Summary

Brings snapshot-streaming `release/integrationnet` onto the v4.1.0 SDK line and fixes the state-proof
bug that took the IntegrationNet indexer down for roughly two hours on 2026-08-10. `sub-trie-roots` is
the only `fields-added-ordinals` gate that changes the **signed** `GlobalSnapshotStateProof`, which this
indexer independently re-derives and validates. IntegrationNet crossed its activation ordinal
**5,880,000** and the indexer then rejected **every** global snapshot from that ordinal on -- `StateProof
Broken`, 16 of 17 fields differing, agreeing only on `mptRoot` -- storing nothing to S3, OpenSearch or
Postgres.

Root cause: `SnapshotProcessor` **hand-built** the proof as an mptRoot-only literal and never consulted
the state-proof selector, so it could only ever produce the not-activated shape.

**Testnet and mainnet will fail identically when their gates activate.** Both are still `9999999`. See
the pre-activation checklist at the bottom.

## Changes

* **Modified** - `SnapshotProcessor.scala`: **the fix.** Replace the hand-built proof literal
  (`Hash.empty` x3, `None` x13, `Some(mptRoot)`) with
  `GlobalSnapshotInfo.assembleMptProof[F](snapshotInfo, snapshot.ordinal, mptRoot.value)`.
  `assembleMptProof`'s else-branch **is** that same literal, so pre-activation behavior is
  byte-identical; post-activation it derives the per-field sub-trie roots from `snapshotInfo`, honoring
  `GlobalStateProofSelector.subTrieRootsEnabled`. `mptRoot` still comes from the pre-built producer trie,
  unchanged.
* **Modified** - `App.scala`: two-argument `GlobalStateProofSelector(lastLegacyOrd, subTrieOrd)`, plus a
  startup line reporting the resolved ordinals and the config map keys. Necessary but on its own inert,
  because the validator never consulted the selector -- see the diagnostic note below.
* **Modified** - `AppS3.scala`: the same two-argument fix. `AppS3` is a **second** `IOApp` entry point
  (the build prints `multiple main classes detected`); correcting only `App` leaves the two able to
  drift.
* **Modified** - `project/Dependencies.scala`: tessellation-sdk `4.1.0-rc.0` -> `4.1.0-rc.4`.
  IntegrationNet moved to rc.4 on 08-07 while the deployed jar was still built against rc.0.
* **Removed** - now-unused `GlobalSnapshotStateProof` and `Hash` imports (scalafix has
  `removeUnused = true`).

Earlier commit on this branch, unchanged from the original PR: the `4.0.0-rc.5 -> 4.1.0-rc.0` bump and
its four SDK adaptations (`CurrencySnapshotValidator.make` arity,
`GlobalSnapshotStateChannelEventsProcessor.make` signature, dropping `forkInfoStorage` from
`nodeSharedConfig`, and the two new `LastSnapshotStorage` members).

## Testing

* `sbt compile` and `sbt assembly` clean against `tessellation-sdk 4.1.0-rc.4` (Maven Central).
* **Verified live on IntegrationNet** (52.53.243.4). After the 16:31:54 UTC restart: zero `StateProof
  Broken`, zero `State broken` warnings, zero errors. Backfill resumed at exactly the activation
  ordinal and validated 5,880,000 through 5,880,031, ending in `Snapshot 5880031 ... sent to postgres`.
* The startup line confirmed configuration was correct all along, which is what isolated the fault:

  ```
  [STATE-PROOF-SELECTOR] env=Integrationnet lastLegacyStateProofOrdinal=5075000
    subTrieRootsActivationOrdinal=5880000
    subTrieRootsKeys=Dev,Integrationnet,Mainnet,Testnet
    lastLegacyKeys=Testnet,Dev,Integrationnet,Mainnet
  ```

  A populated acceptance trace next to an **empty** validator proof at the same ordinal means two code
  paths in the consumer, not two configurations. `println` rather than `logger.info` because these are
  plain vals outside the `IO` chain, so an unsequenced `logger.info` would never run; stdout is captured
  by journald.
* **No unit test added.** Nothing in this repo currently catches a hand-built proof drifting from the
  SDK's assembly. A regression test asserting the validator's proof equals `assembleMptProof` output for
  an ordinal at or after an activation ordinal is the obvious follow-up, and worth doing before the
  mainnet gate.
* **CI note:** `pull_request.yml` is stale and cannot pass on any 4.1.x pin. It targets the retired
  `ubuntu-20.04` runner label, so the job never gets scheduled -- the run on this PR sat queued exactly
  24h and was then `CANCELLED`, before these commits. Even if it were scheduled, it checks out
  tessellation `v2.12.0` and sets up Java 11, while the 4.1.x SDK requires Java 21. Worth a separate PR
  to modernise; deliberately out of scope for a hotfix.

## Pre-activation checklist for testnet and mainnet

Both nets still carry `sub-trie-roots: 9999999`. When either is scheduled to activate:

1. Deploy a snapshot-streaming jar containing **all three** source changes above **before** the gate
   ordinal, in the same coordinated cold restart as the node software.
2. Confirm the startup line reads `subTrieRootsActivationOrdinal=<that net's gate>`, not
   `9223372036854775807` (`MaxValue`, the silent default that means OFF).
3. Diff **every** `fields-added-ordinals` gate on the box against the tessellation release. The box's own
   `application.conf` cannot inherit SDK defaults: Typesafe Config auto-loads `reference.conf` from every
   jar but only one `application.conf`, and the SDK ships its gate defaults in `application.conf`, so
   each consumer must restate them.
4. Only then let the chain cross the ordinal.

Longer-term, moving those gate defaults to `reference.conf` in the SDK would remove step 3 entirely.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -20,7 +20,7 @@ object Dependencies {
     val logstash = "7.2"
     val organizeImports = "0.6.0"
     val refined = "0.10.1"
-    val tessellation = "4.0.0-rc.5"
+    val tessellation = "4.1.0-rc.0"
     val weaver = "0.8.1"
   }
 
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -20,7 +20,7 @@ object Dependencies {
     val logstash = "7.2"
     val organizeImports = "0.6.0"
     val refined = "0.10.1"
-    val tessellation = "4.1.0-rc.0"
+    val tessellation = "4.1.0-rc.4"
     val weaver = "0.8.1"
   }
 
--- a/src/main/scala/org/constellation/snapshotstreaming/App.scala
+++ b/src/main/scala/org/constellation/snapshotstreaming/App.scala
@@ -31,8 +31,30 @@ object App extends IOApp {
             KryoSerializer.forAsync[IO](shared.sharedKryoRegistrar).use { implicit ks =>
               JsonSerializer.forAsync[IO].asResource.use { implicit jsonSerializer =>
                 val hashSelect = makeHashSelect(appConfig, sharedCfg)
+                // `subTrieRootsActivationOrdinal` MUST be passed. It defaults to MaxValue (sub-trie
+                // roots OFF), and `sub-trie-roots` is the only fields-added gate that changes the
+                // SIGNED GlobalSnapshotStateProof, which this indexer independently re-derives and
+                // validates. With gl0 signing roots ON and this selector OFF, every post-activation
+                // ordinal fails `StateProof Broken` on the sub-trie fields and nothing is indexed.
+                // Mirrors gl0's own construction in TessellationIOApp.
+                // Resolved into named vals so the values LOGGED are byte-identical to the values
+                // USED -- a diagnostic that recomputes them can disagree with the selector.
+                val ssEnv = appConfig.snapshotStreaming.environment
+                val lastLegacyOrd =
+                  sharedCfg.lastLegacyStateProofOrdinal.getOrElse(ssEnv, SnapshotOrdinal.MinValue)
+                val subTrieOrd =
+                  sharedCfg.fieldsAddedOrdinals.subTrieRoots.getOrElse(ssEnv, SnapshotOrdinal.MaxValue)
+                // println, not the IO logger: these are plain vals outside the IO chain, so an
+                // unsequenced logger.info would never run. stdout is captured by journald.
+                println(
+                  s"[STATE-PROOF-SELECTOR] env=$ssEnv " +
+                    s"lastLegacyStateProofOrdinal=${lastLegacyOrd.value.value} " +
+                    s"subTrieRootsActivationOrdinal=${subTrieOrd.value.value} " +
+                    s"subTrieRootsKeys=${sharedCfg.fieldsAddedOrdinals.subTrieRoots.keys.mkString(",")} " +
+                    s"lastLegacyKeys=${sharedCfg.lastLegacyStateProofOrdinal.keys.mkString(",")}"
+                )
                 implicit val gsps: GlobalStateProofSelector =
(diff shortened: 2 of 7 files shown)


ryle-ai (commented):
Reviewed the 4.1.0-rc.4 bump and the state-proof validation rework across the two entry points, the snapshot processor, and the service wiring.

The proof path now assembles the SDK's full 17-field MPT proof instead of the hand-built mptRoot-only literal, so post-activation snapshots (where gl0 signs per-field sub-trie roots into fields that were previously None) validate instead of breaking the chain. The two ordinal gates are distinct and correctly composed: the processor's outer `snapshot.ordinal > lastLegacyStateProofOrdinal` selects legacy-validate vs MPT-assemble, while the selector's inner `subTrieRoots` activation ordinal selects mptRoot-only vs per-field sub-trie roots — different ordinals, not conflated. Both resolve per-environment via getOrElse with MinValue/MaxValue defaults, so an unlisted environment fails safe toward legacy-validate / mptRoot-only.

Implicits check out: `assembleMptProof` runs inside the enclosing `HasherSelector[F].forOrdinal` block that supplies `Hasher[F]`, and the two-arg `GlobalStateProofSelector` / `CurrencyStateProofSelector.instance` are wired identically in App and AppS3 (kept in lockstep so the entry points can't drift). The service reconciliation matches the SDK's rc.4 shape — four-arg `CurrencySnapshotValidator.make` (drops tessellation3Migration) and the new `mptStore`/`fieldsAddedOrdinals`/`env` args on `GlobalSnapshotStateChannelEventsProcessor.make` — and the `forkInfoStorage` positional removal lines up with the new `SharedConfig`.

Considered and set aside: `setForRecovery` and `clear` on the file-based storage are added but have no call site yet (recovery-path infrastructure), so they don't affect the indexing path this PR fixes.

